Issue #1, opened 2024-05-09
Enable precompiled headers in docker image


WardBrian:
As of https://github.com/WardBrian/tinystan/commit/6688eb70e18b0e6911e4fac3988309bfc5c2ec73, the following added to `make/local` should result in faster compile times:

```makefile
CXX_TYPE=clang # Stan's auto-detection isn't smart enough for em++
PRECOMPILED_HEADERS=true
```

magland:
@WardBrian 

I confirm that the compilation time has decreased from 18 seconds down to 13 seconds for the basic example.

The new version has been pushed to dockerhub, but it is not yet loaded on the public server. I'll do that soon.

magland:
New image now loaded on public server, and it reliably gives 13.5 seconds for basic example

WardBrian:
Some other tricks we might want to try are not passing `-g`, and setting `-DCMAKE_BUILD_TYPE=Release` in the TBB cmake step (the default is RelWithDebug, which builds larger objects that might be slower to process downstream)

Both of these things would also make the generated code less debuggable, so they should probably be done much later in the lifecycle.

PR #2, opened 2024-05-09
update dockerfile

magland:
Implements #1 

@WardBrian I pushed the new docker image, but I didn't particularly notice a speedup in compilation.


WardBrian:
Hi @magland - I think the end of line comment was making the `CXX_TYPE` equal to `clang ` (note the trailing space), which was the issue. 

Now if you re-build you'll see the command to build the `main.o` object includes the flag `-include-pch /some/path/here`, and it does seem to be noticeably faster.

PR #5, opened 2024-05-14
Move sampling into a Web Worker

### LINKED ISSUE #3: non-blocking sampling - need report progress and enable canceling
@WardBrian 

Would be great if we could expose a lower level entry point into tinystan so that we can perform sampling in a non-blocking manner and provide real-time updates to the user. Right now the full sampling is done in a single blocking call to `_tinystan_sample`

Alternatively could we get a callback from _tinystan_sample with the progress? And perhaps if the callback returns false, then it cancels the sampler?

Would be great to be able to hook into mcmc monitor.

WardBrian:
This closes #3.

[Video preview](https://github.com/flatironinstitute/stan-playground/assets/31640292/128027b9-2c80-4b3f-a305-3f9d8ce9104b)

(it's not that choppy in reality, my screen recorder has some artifacts)

Overview of changes:

- Web workers support doing work out of the main thread _and_ termination. However, this termination leaves the worker 'dead', and so it needs to be re-created. It also works in its own scope, and communicates only through messages.
- As a result, the home page no longer stores a Stan model. Instead, the state we care about is the URL of the compiled Stan model
   - StanFileEditor.tsx was changed to set this, rather than doing any loading
- The RunPanel now stores a worker. This is created whenever the URL changes, which also kills the previous worker. The same logic is re-used for cancellation.
- The worker can receive two kinds of message: 
   - `load`, which provides the url to import
   - `sample`, which provides a config object for the sampler (data, seed, etc)
- The worker can reply with three kinds of message:
   - `modelLoaded`, sent after a `load` message is received and successfully handled 
   - `sampleReturn`, sent after the sampler completed. Always has an `error` payload. If `error === null`, then it also has `draws`.
   - `progress`, sent during sampling based on the strings the Stan sampler prints at regular intervals.


magland:
Terrific!

PR #6, opened 2024-05-14
Also provide parameter names in return

WardBrian:
This allows us to do downstream processing using the variable names of the user's model. As an example I've now updated the output area to show the means of each parameter above the full draws.


gui/src/app/exampleStanies/exampleStanies.ts
```diff
@@ -26,11 +26,36 @@ model {
     y ~ normal(alpha + beta * x, sigma);
 }
     `.trim()
+
+// generated with beta = 1.25, alpha = -0.1, sigma = 0.1
```
WardBrian, line 30: The ability to see the means made me realize why this simple linear regression was so slow. The data we were providing was perfectly noiseless, which means sigma was being driven down to zero. But the constraints in stan like `lower=0` are _exclusive_, so you can get 'bouncing' behavior in these cases which slows down HMC.

Adding in a bit of noise means this model now runs very quickly (perhaps _too_ quickly to adequately display the progress/cancellation features -- we can add harder models, if we want that)

gui/src/app/RunPanel/RunPanel.tsx
```diff
@@ -188,4 +188,31 @@ const LinearProgressWithLabel = (props: LinearProgressProps & { value: number })
...
+const DrawsDisplay: FunctionComponent<DrawsDisplayProps> = ({ draws, paramNames }) => {
+    const means: { [k: string]: number } = {};
+
+    for (const [i, element] of paramNames.entries()) {
+        let sum = 0;
+        for (const draw of draws[i]) {
+            sum += draw;
+        }
+        means[element] = sum / draws[i].length;
+    }
```
jsoules, line 206: Other than my temptation to replace this with an inscrutable one-liner, this looks good on inspection. Will try kicking the tires in a sec.

gui/src/app/RunPanel/RunPanel.tsx
```diff
@@ -24,6 +24,8 @@ const RunPanel: FunctionComponent<RunPanelProps> = ({ width, height, compiledUrl
     const [runStatus, setRunStatus] = useState<RunStatus>('waiting');
     const [errorMessage, setErrorMessage] = useState<string>('');
     const [samples, setSamples] = useState<number[][] | undefined>(undefined);
+    const [paramNames, setParamNames] = useState<string[] | undefined>(undefined);
```
jsoules, line 27: Apart from conditioning run doneness on `paramNames` being non-undefined (l. 147), any reason not to keep this as a simple `string[]` and use an empty list for the default?

gui/src/app/RunPanel/RunPanel.tsx
jsoules: I'm assuming there's no valid condition where one chain would have 0 draws. (Or--at that point something has already blown up so we don't need to worry about the divide-by-zero, perhaps?)

gui/src/app/tinystan/index.ts
```diff
@@ -39,6 +39,17 @@ interface WasmModule {
...
+
+const HMC_SAMPLER_VARIABLES = [
+  "lp__",
+  "accept_stat__",
+  "stepsize__",
+  "treedepth__",
+  "n_leapfrog__",
+  "divergent__",
+  "energy__",
+];
```
jsoules, line 51: We don't expose anything through this interface that would cause it to use a different sampler with different default variables, right?

gui/src/app/tinystan/index.ts
```diff
@@ -164,7 +175,7 @@ export default class StanModel {
   // - inits
   // - init inv metric
   // - save_metric
-  public sample(p: Partial<SamplerParams>): number[][] {
+  public sample(p: Partial<SamplerParams>): { paramNames: string[], draws: number[][] } {
```
jsoules, line 178: Consider naming this type? Seems like it might be important.

gui/src/app/exampleStanies/exampleStanies.ts
jsoules: Nice catch. This does run extremely fast now. Aw shucks, have to use a more powerful example... :smile:

jsoules (approved):
Yeah, this all looks good. I made a couple minor notes--take or leave as you will--and will ping you about a couple other things, but I'd merge this.

gui/src/app/tinystan/index.ts
WardBrian: Nope. If we want to do e.g. Pathfinder later we would need a different list of these algorithmic parameters, but we'd also have that behind a different function than `.sample()`

gui/src/app/RunPanel/RunPanel.tsx
WardBrian: No, and we could probably do the same for `samples`.

gui/src/app/RunPanel/RunPanel.tsx
WardBrian: Sample has a precondition that the number of draws is at least 1, and a successful return means we have all the requested draws, so we don't need to worry here

PR #7, opened 2024-05-14
Add SIR example model

WardBrian:
This adds a much more significant example model which is modeling disease transmission. The basic idea is that people are either susceptible, infected (and infectious), or recovered (and immune). A similar model is well-explained here https://mc-stan.org/users/documentation/case-studies/boarding_school_case_study.html

This model is cool because it shows off the ability of Stan to use an ODE solver as part of your model likelihood. 

With the default settings it takes ~30-40 seconds per chain, so it is definitely a decent showing of the progress bar and cancellation after #6.

This model also motivates the ability to control the sampler's arguments (typically we only run it for ~100 iterations per chain, not ~1000), and also possibly the initial values (a json similar to data)

### DIFF
--- a/gui/src/app/exampleStanies/exampleStanies.ts
+++ b/gui/src/app/exampleStanies/exampleStanies.ts
@@ -66,4 +66,76 @@ examplesStanies.push({
 })
 
 
+const SIRStan = `
+// the "susceptible-infected-recovered" model
+functions {
+  vector sir(real t, vector y, real beta, real gamma, int N) {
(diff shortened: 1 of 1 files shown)


jsoules:
Confirming that this runs (although slowly) on my machine as well.
Very interesting example, and suggests a couple things that we could/should do to enable user-extensible example sets and expose more configuration.

magland:
Confirmed that it runs.

PR #10, opened 2024-05-15
StanSampler and misc refactor

magland:
@WardBrian @jsoules 

Description of changes in this PR

(this branch has been deployed here: https://stan-playground.vercel.app/)

* New StanSampler class that manages the worker as an internal variable and also manages the state and the sampling outputs
* Lifted the sampler state up to the HomePage (so that it is not just confined to the RunPanel)
* Split the RunPanel into two parts: RunPanel and SamplerOutputView
* SamplerOutputView is a tab widget that shows a Summary tab and a Draws tab.
* The Summary tab shows the means of the variables
* The Draws tab shows a table of all the draws
* There is an export to .csv button on the draws table
* Some adjustments on the layout (default sizes of panels, etc)

### DIFF
--- a/gui/src/app/RunPanel/RunPanel.tsx
+++ b/gui/src/app/RunPanel/RunPanel.tsx
@@ -97,7 +97,16 @@ const SamplingProgressComponent: FunctionComponent<SamplingProgressComponentProp
     return (
         <>
             <div style={{ width: "60%" }}>
-                <LinearProgressWithLabel sx={{ height: 10 }} value={progress} />
+                <LinearProgressWithLabel
+                    sx={{
+                        height: 10,
+                        // https://stackoverflow.com/a/73009519
+                        "& .MuiLinearProgress-bar": {
+                            transition: 'none'
+                        }
+                    }}
+                    value={progress}
+                />
             </div>
             <div>
                 Chain {report.chain} Iteration: {report.iteration} / {report.totalIterations} ({report.warmup ? 'Warmup' : 'Sampling'})
--- a/gui/src/app/RunPanel/RunPanel.tsx
+++ b/gui/src/app/RunPanel/RunPanel.tsx
@@ -1,153 +1,113 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { FunctionComponent, useCallback, useEffect, useState } from 'react';
+import Box from '@mui/material/Box';
 import LinearProgress, { LinearProgressProps } from '@mui/material/LinearProgress';
 import Typography from '@mui/material/Typography';
-import Box from '@mui/material/Box';
+import { FunctionComponent, useCallback, useEffect, useState } from 'react';
 
-import StanWorker from '../tinystan/Worker?worker';
-import { Progress, Replies, Requests } from '../tinystan/Worker'
+import StanSampler, { StanSamplerStatus } from '../StanSampler/StanSampler';
+import { Progress } from '../tinystan/Worker';
 
 type RunPanelProps = {
     width: number;
     height: number;
-    compiledUrl: string;
+    sampler?: StanSampler;
     data: any | undefined
     dataIsSaved: boolean
 };
 
-type RunStatus = '' | 'waiting' | 'running' | 'done' | 'failed';
+const numChains = 4;
 
-const chains = 4;
+const RunPanel: FunctionComponent<RunPanelProps> = ({ width, height, sampler, data, dataIsSaved }) => {
 
-const RunPanel: FunctionComponent<RunPanelProps> = ({ width, height, compiledUrl, data, dataIsSaved }) => {
-
-    const [runStatus, setRunStatus] = useState<RunStatus>('waiting');
+    const [runStatus, setRunStatus] = useState<StanSamplerStatus>('');
     const [errorMessage, setErrorMessage] = useState<string>('');
-    const [samples, setSamples] = useState<number[][]>([[]]);
-    const [paramNames, setParamNames] = useState<string[]>([]);
 
     const [progress, setProgress] = useState<Progress | undefined>(undefined);
 
-    const [modelWorker, setModelWorker] = useState<Worker | undefined>(undefined);
-
-    // Cancellation destroys the worker, and therefore
-    // requires the same code to be run as if a new
-    // URL was provided. This state is therefore
-    // used to force the following useEffect hook to rerun.
-    const [trigger, setTrigger] = useState(false);
-
     useEffect(() => {
-        if (!compiledUrl) {
-            setModelWorker(undefined);
-            return;
-        }
-        const worker = new StanWorker();
-        setModelWorker(worker);
-        worker.postMessage({ purpose: Requests.Load, url: compiledUrl });
-        return () => {
-            if (worker) {
-                console.log("Cleaning up worker");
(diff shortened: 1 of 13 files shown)


gui/src/app/StanSampler/StanSampler.ts
```diff
@@ -0,0 +1,85 @@
...
+                        this.#status = 'completed';
+                        this.#onStatusChangedCallbacks.forEach(cb => cb())
+                    }
+                    break;
+                }
+            }
+        }
+        this.#worker.postMessage({ purpose: Requests.Load, url: this.compiledUrl });
+    }
+    sample(a: {data: any, numChains: number}) {
```
WardBrian, line 49: The right type to accept here would be `SamplerParams` from TinyStan/index.ts and just pass it unchanged, I think

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -93,6 +95,20 @@ const HomePage: FunctionComponent<Props> = ({ width, height }) => {
...
+
+    useEffect(() => {
+        if (!compileMainJsUrl) {
+            setSampler(undefined);
+            return;
+        }
+        const s = new StanSampler(compileMainJsUrl);
+        setSampler(s);
+        return () => {
+            s.cancel();
```
WardBrian, line 108: I don't think this will work, since `s.cancel()` will end up re-calling initialize?

Have you verified there aren't dangling threads after this?

Edit: there do appear to be in the firefox debugger

gui/src/app/RunPanel/RunPanel.tsx
```diff
@@ -1,153 +1,113 @@
...
-    }, [compiledUrl, trigger])
+        })
+        sampler.onProgress((progress: Progress) => {
+            if (canceled) return;
+            setProgress(progress);
+        })
+        return (
+            () => {
+                canceled = true;
+                sampler.cancel();
```
WardBrian, line 44: Is the `canceled` variable really necessary? After `terminate` is called I wouldn't expect any further callbacks to fire

WardBrian (commented):
The newer output display and the save to CSV both look great!

I'm not entirely sold on the StanSampler class. I think it may be a bit premature to build an abstraction around that idea, and it seems like it leads to a lot of extra callbacks elsewhere

gui/src/app/pages/HomePage/HomePage.tsx
magland: You're right. I changed it to terminate. Now the class has two methods: cancel() and terminate(). The latter does not restart the worker thread.

gui/src/app/RunPanel/RunPanel.tsx
magland: I think it's safest to keep this in here because you could conceivably have a situation where the sampler prop changes, but the old sampler is still alive and emitting  callbacks (doesn't actually happen in the current program, but we want this RunPanel component to be independent). The `canceled` here doesn't mean that the sampler was canceled, but it means that the useEffect was canceled due to the sampler prop changing.

gui/src/app/StanSampler/StanSampler.ts
magland: I have made this change

magland: Ooops, the name of my commit message was "i have made this change" when that was supposed to be my comment in this issue. Please forgive. :)

magland:
> I'm not entirely sold on the StanSampler class. I think it may be a bit premature to build an abstraction around that idea, and it seems like it leads to a lot of extra callbacks elsewhere

@WardBrian The reason I did this was so that I could lift the state up to the top level component... and encapsulate sampling state, output, and worker communication in one entity that could be passed down to the child components which can interact with it. I'm not opposed to refactoring this again in the future to reduce complexity... but shall we keep it in for now? Unless you want to take a crack at achieving the same thing in a less heavy way...

gui/src/app/StanSampler/StanSampler.ts
WardBrian: I think I misled you, it should be `Partial<SamplerParams>`, which lets `defaultSamplerParams` stay private (which it would be if/when `tinystan.ts` ended up in its own package

gui/src/app/pages/HomePage/HomePage.tsx
WardBrian: I am still seeing the number of threads grow after each compilation

magland: I made some adjustments to StanSampler and now this should be fixed. The problem was that an additional call to cancel() was being made which reinitialized the worker. Now cancel() only has any effect if the status was 'sampling'.

WardBrian:
The compile button could stay where it is if the home page still stores the URL as it does now

magland:
> The compile button could stay where it is if the home page still stores the URL as it does now

Okay good point... working on it...

gui/src/app/pages/HomePage/HomePage.tsx
WardBrian: I can confirm there is no extra threads now, but I am getting the "Nothing to cancel" warning, so I think there is still a call being made that shouldn't be there, it's just been turned into a no-op now?

magland: Thanks, I have now removed that extra no-op call.

magland:
@WardBrian The StanSampler state has now been lowered back down to the `LowerRightView` which passes sampler as a prop to RunPanel and SamplerOutputView.

I tried moving away from using a class and having something simpler like worker/status as you suggested, but it was getting pretty complicated. I moved all the callback logic into `StanSampler/useStanSampler.ts` with custom hooks... and now the usage is pretty straightforward in the RunPanel and SamplerOutputView components.

If you still have concerns maybe we could chat in person.

jsoules:
I think moving into the hooks model in your most recent commit is a good move--what I'd been writing (before lunch) was that the extensive callback structure, while flexible, winds up making the consumers (`RunPanel`, `SamplerOutputView`) need to know a fair amount about the internal state of the `StanSampler`, at which point the abstraction isn't necessarily making it easier to use. However, especially given the custom hooks, I don't think the full callback structure is still needed--it's probably easier just to change the interface object's state directly in response to worker-thread messages.

I think in general that's where the biggest utility win of a `StanSampler` would lie: providing a clear interface for worker-thread communication, with data validation on the way in. I'd like to have a quick chat about the full scope of functionality of the `StanSampler` class--it seems like we might be better served if the status logic lives on the `Worker` object itself, for instance, especially if it impacts the behavior of the `Worker`.

I'll do some line-by-line review as well, but I don't want to chase a moving target too much so I'll defer until after we're all on the same page.

magland:
> I'm now seeing some weird behavior with the progress bar:

@WardBrian This has been fixed. It was a MaterialUI styling issue.

magland:
Shall we merge this.... or do you guys want to do any further polishing?

@WardBrian @jsoules

WardBrian:
I think this is good as it stands.

I have a few small things I might tweak the next time I touch the files, and I agree with @jsoules comment/our discussion right now that there are some questions if the object could be 'skinnier', but I like where it sits.

PR #14, opened 2024-05-16
Update tinystan.ts to support initialization

WardBrian:
This updates the `sample` method in `tinystan/index.ts` to include all the arguments supported by the tinystan bindings in other languages. The most relevant of these is the ability to supply initial values of parameters with `inits`.

The `init_inv_metric` argument is currently unused, but I've included it such that the type of `SamplerParams` should be pretty stable going forward.

### DIFF
--- a/gui/src/app/tinystan/index.ts
+++ b/gui/src/app/tinystan/index.ts
@@ -61,18 +61,22 @@ export type StanDraws = {
   draws: number[][];
 };
 
+export type StanVariableInputs = {
+  [k: string]: unknown;
+};
+
 export interface SamplerParams {
-  data: string | object;
+  data: string | StanVariableInputs;
   num_chains: number;
-  inits: string | object | string[] | object[];
+  inits: string | StanVariableInputs | string[] | StanVariableInputs[];
   seed: number | null;
   id: number;
   init_radius: number;
   num_warmup: number;
   num_samples: number;
   metric: HMCMetric;
-  save_metric: boolean,
-  init_inv_metric: number[] | number[][] | null;
+  save_metric: boolean;
+  init_inv_metric: number[] | number[][] | number[][][] | null;
   adapt: boolean;
   delta: number;
   gamma: number;
@@ -113,7 +117,7 @@ const defaultSamplerParams: SamplerParams = {
   stepsize: 1.0,
   stepsize_jitter: 0.0,
   max_depth: 10,
-  refresh: 100,  // this is how often it prints out progress
+  refresh: 100,
   num_threads: -1,
 };
 
@@ -160,9 +164,11 @@ export default class StanModel {
 
   private encodeInits(inits: string | object | string[] | object[]): cstr {
     if (Array.isArray(inits)) {
-      return this.encodeString(inits.map((i) => jsonify(i)).join(this.sep));
+      return this.encodeString(
+        inits.map(i => string_safe_jsonify(i)).join(this.sep),
+      );
     } else {
-      return this.encodeString(jsonify(inits));
+      return this.encodeString(string_safe_jsonify(inits));
     }
   }
 
@@ -171,7 +177,7 @@ export default class StanModel {
     seed: number,
     f: (model: model_ptr) => T,
   ): T {
-    const data_ptr = this.encodeString(jsonify(data));
+    const data_ptr = this.encodeString(string_safe_jsonify(data));
     const err_ptr = this.m._malloc(4);
     const model = this.m._tinystan_create_model(data_ptr, seed, err_ptr);
     this.m._free(data_ptr);
@@ -235,7 +241,7 @@ export default class StanModel {
       const rawParamNames = this.m.UTF8ToString(
         this.m._tinystan_model_param_names(model),
       );
-      const paramNames = HMC_SAMPLER_VARIABLES.concat(rawParamNames.split(','));
+      const paramNames = HMC_SAMPLER_VARIABLES.concat(rawParamNames.split(","));
 
       const n_params = paramNames.length;
 
@@ -250,9 +256,16 @@ export default class StanModel {
       let metric_out = NULL;
       if (save_metric) {
         if (metric === HMCMetric.DENSE)
-          metric_out = this.m._malloc(free_params * free_params * Float64Array.BYTES_PER_ELEMENT);
+          metric_out = this.m._malloc(
+            num_chains *
+              free_params *
+              free_params *
+              Float64Array.BYTES_PER_ELEMENT,
+          );
         else
-          metric_out = this.m._malloc(free_params * Float64Array.BYTES_PER_ELEMENT);
+          metric_out = this.m._malloc(
+            num_chains * free_params * Float64Array.BYTES_PER_ELEMENT,
+          );
       }
 
       // Allocate memory for the output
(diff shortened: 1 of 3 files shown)


gui/src/app/tinystan/index.ts
```diff
@@ -280,17 +305,35 @@ export default class StanModel {
...
       // copy out parameters of interest
-      const draws: number[][] = Array.from({ length: n_params }, () => []);
-      for (let i = 0; i < n_draws; i++) {
-        for (let j = 0; j < n_params; j++) {
-          const elm = out_buffer[i * n_params + j];
-          draws[j][i] = elm;
-        }
-      }
+      const draws: number[][] = Array.from({ length: n_params }, (_, i) =>
+        Array.from({ length: n_draws }, (_, j) => out_buffer[i + n_params * j])
```
jsoules, line 309: This went from "multiply by `i` and add `j`" to the other way around. Looks like `i` is still the outer index and `j` the inner--what's up? Did the memory layout change?

gui/src/app/tinystan/index.ts
```diff
@@ -280,17 +305,35 @@ export default class StanModel {
...
+      let metric_array: number[] | number[][] | null = null;
+
+      if (save_metric) {
+        if (metric === HMCMetric.DENSE) {
+          const metric_buffer = this.m.HEAPF64.subarray(
+            metric_out / Float64Array.BYTES_PER_ELEMENT,
+            metric_out / Float64Array.BYTES_PER_ELEMENT + free_params * free_params,
+          );
+          metric_array = Array.from({ length: free_params }, (_, i) =>
+            Array.from({ length: free_params }, (_, j) => metric_buffer[i * free_params + j])
```
jsoules, line 324: And here it's "times i plus j"

gui/src/app/tinystan/index.ts
```diff
@@ -311,3 +354,12 @@ export default class StanModel {
     return version;
   }
 }
+
+
+const jsonify = (obj: string | object): string => {
```
jsoules, line 359: I'd consider calling it `string_safe_jsonify` to clarify why you aren't just using `JSON.stringify()` directly (but it's probably also fine in the file-local context)

gui/src/app/tinystan/index.ts
```diff
@@ -151,17 +158,21 @@ export default class StanModel {
     throw new Error(err_msg);
   }
 
+  private encodeInits(inits: string | object | string[] | object[]): cstr {
+    if (Array.isArray(inits)) {
+      return this.encodeString(inits.map((i) => jsonify(i)).join(this.sep));
+    } else {
+      return this.encodeString(jsonify(inits));
+    }
+  }
```
jsoules, line 167: I'm missing enough context to be clear--does this do the right thing if `inits` is an object (that might contain entries that would otherwise be multiple JSON values)? What are the circumstances under which it might be?

Maybe not possible at this point but it might be nice to give some type hinting for what this could be.

jsoules (commented):
Just a couple questions.

gui/src/app/tinystan/index.ts
WardBrian: The memory layout didn't change, just the previous way of writing it was a bit opaque -- before, `i` was from 0 to `num_draws`, and `j` went from 0 to `num_params`, which is why the assign in to draws was `[j][i]` -- flipped from how you would expect.

In addition to being a bit more compact, this style of just directly using `Array.from` forced me to actually write my indexes in the order they should be

gui/src/app/tinystan/index.ts
WardBrian: I'm not able to follow what you're asking for, could you elaborate?

jsoules: Sure. So there are two cases here:

- `inits` is an `Array` (presumably of strings) -- then we encode each string (array element) in json and join them with the separator
- `inits` is not an `Array`--then we just encode it as one unit

But the typing suggests it could also be an (otherwise undetermined) `object`. I'm asking what potential use cases there would be when an object is passed, and would this logic cover those. Can we say anything about the shapes of valid object inputs here?

As a concrete example, `let isObj = [1, 2, 3]; typeof(isObj)` is 'object'. But `Array.isArray(isObj)` is also true, so we'd be JSONifying the individual values as separate JSON objects--which I assume isn't really what we want?

Conversely, if `inits` is `[[1, 2, 3], [4, 5, 6]]`, I expect this to return `[1, 2, 3] %SEP% [4, 5, 6]`, is that the correct behavior?

WardBrian: Ah I follow now.

There are two "output behaviors" here, which is what we branch on:

1. We output a string containing a single json object (every chain gets this same initial value)
2. We output a string containing several json objects separated by a special character (different chains get different initial values)

The json objects in question will look like, in the simplest case, `{theta: 0.2}`. This corresponds to initializing the parameter called theta at 0.2. The values can be more complicated if you have e.g. a matrix parameter

As such, `[[1,2,3], [4,5,6]]` is invalid from the jump. I think until now I didn't realize the `object` type meant something like `Object` in Java - I was assuming it meant `object` in the JSON sense of key-value pairs.

If I change the type to 

```ts
type StanObject /* open to name suggestions */ = { [k:string]: any }
encodeInits(inits: string |StanObject | string[] | StanObject[])
```
is that more correct?

gui/src/app/tinystan/index.ts
jsoules: If it's performance-impactful to the level that row-major/column-major choice will be noticeable, you may actually want something from either the [math.js](https://mathjs.org/) library or an [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) object.

But isn't this pretty much just getting passed to the sampler worker, which will copy it (once) into memory managed by the WASM code?

WardBrian: This is output, so it really depends on what we end up doing with it later. The input equivalent to this is the `init_inv_metric` argument that I've left unimplemented for now, and it's certainly true that something like math.js would be helpful there

gui/src/app/tinystan/index.ts
jsoules: Ahh, I get it. So we're always going to be using the name of the parameter plus its initial value.

I would probably use `StanParameterValue` instead of `StanObject` to be more semantic about it.

I think what you're really looking for is a [Record<string, number>](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) or list of same, so maybe:

```typescript
type StanParameterValue = Record<string, number>
encodeInits(inits: string | StanParameterValue | string[] | StanParameterValue[])
```
?

gui/src/app/tinystan/index.ts
jsoules: Cool. Sounds like something we can worry about once it's used for something other than just display.

gui/src/app/tinystan/index.ts
WardBrian: Two wrinkles -

1. I had already made this mistake before, this new type is also what `data` really is (which currently uses `object`, due to my same confusion). So the name should be a bit more generic
2. Record sounds good (and equivalent?), but `number` is too narrow.

Consider the stan model

```stan
parameters {
  vector[3] sigma;
}
```

A valid JSON is
```js
{
  "sigma": [1, 2, 3]
}
```

This can get arbitrarily complicated with tuple types in Stan now

```stan
tuple(real, matrix[2, 3] x;
```

would have a JSON value
```json
"x":  { "1" : 3.14, "2": [[1,2], [3,4], [5,6]]}
```

jsoules: > this new type is also what `data` really is

:+1:

> `number` is too narrow

Ah, I was afraid of that. Well, so long as the keys are serializable I suppose! If there's really no logic we need to apply here, we can just treat it as a blob.

jsoules (approved):
This seems reasonable to me.

PR #17, opened 2024-05-16
Stop leaking memory if sampling fails

WardBrian:
This fixes an issue where exceptions thrown in the main part of the sample function would lead to dangling memory that never got cleaned up.

I already had something like this for the `model` pointer itself, so this just extends that idea to include other generic pointers we need to call `free` on. It ends up looking a little bit like golang or zig's `defer` keyword


jsoules (commented):
Haven't tested this, but you say you have and it all looks reasonable, so go for it.

Issue #16, opened 2024-05-16
Control parameters for sampling


WardBrian:
The following would be nice to have control over:

- [ ]Number of chains (default 4, in practice I think this being a slider from 1 to 8 is probably reasonable, but in theory it can be unbounded)
- [ ] Number of warmup iterations (default 1000, can be [0, $\infty$)) 
- [ ] Number of sampling iterations (default 1000, can be [1, $\infty$))
- [ ] Initializations. This is tricky, can either be 1 json object or a list of num_chains JSON objects.
- [ ] "Initialization radius" (parameters not given an initial value directly are drawn from `uniform(-R, R)` on the unconstrained scale) (default 2.0, can be [0, $\infty$), but in practice limiting to say 10 in the UI is probably fine)
- [ ] Seed. Any uint32.

There are also these, which are more advanced and might belong on a drop down/panel that is hidden by default

- [ ] Metric - decide which form of mass matrix to use. (Default is Diagonal. The set {Diagonal, Dense, Unit})
- [ ] save_warmup: whether the warmup draws are also returned in the output.
- [ ] stepsize: initial stepsize for the sampler
- [ ] adapt: (bool, whether to enable adaptation)
- [ ] delta: target acceptance probability for the dual averaging adaptation
- [ ] max_depth: maximum tree depth allowed for the sampler (# of evals can be on the order of 2^{max_depth}, so be careful)

The rest are probably not priorities, and if they are supported should be grouped with the more advanced options.

magland:
I assigned this to myself.

PR #18, opened 2024-05-16
Add sampling options panel and update sampler to use new options

magland:
Partially addresses #16

### DIFF
--- a/gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
+++ b/gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
@@ -34,56 +34,61 @@ const SamplingOptsPanel: FunctionComponent<SamplingOptsPanelProps> = ({ sampling
                     <tr>
                         <td># chains</td>
                         <td>
-                            <IntEdit
+                            <NumberEdit
                                 value={num_chains}
-                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, num_chains: value })}
+                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, num_chains: value as number })}
                                 min={1}
                                 max={8}
                                 readOnly={readOnly}
+                                type="int"
                             />
                         </td>
                     </tr>
                     <tr>
                         <td># warmup</td>
                         <td>
-                            <IntEdit
+                            <NumberEdit
                                 value={samplingOpts.num_warmup}
-                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, num_warmup: value })}
+                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, num_warmup: value as number })}
                                 min={0}
                                 readOnly={readOnly}
+                                type="int"
                             />
                         </td>
                     </tr>
                     <tr>
                         <td># samples</td>
                         <td>
-                            <IntEdit
+                            <NumberEdit
                                 value={samplingOpts.num_samples}
-                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, num_samples: value })}
+                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, num_samples: value as number })}
                                 min={1}
                                 readOnly={readOnly}
+                                type="int"
                             />
                         </td>
                     </tr>
                     <tr>
                         <td>init radius</td>
                         <td>
-                            <FloatEdit
+                            <NumberEdit
                                 value={samplingOpts.init_radius}
-                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, init_radius: value })}
+                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, init_radius: value as number })}
                                 min={0}
                                 readOnly={readOnly}
+                                type="float"
                             />
                         </td>
                     </tr>
                     <tr>
                         <td>seed</td>
                         <td>
-                            <IntOrUndefinedEdit
+                            <NumberEdit
                                 value={samplingOpts.seed}
                                 onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, seed: value })}
                                 min={0}
                                 readOnly={readOnly}
+                                type="int"
                             />
                         </td>
                     </tr>
@@ -96,67 +101,42 @@ const SamplingOptsPanel: FunctionComponent<SamplingOptsPanelProps> = ({ sampling
     )
 }
 
-type InitEditProps = {
-    value: number;
-    onChange: (value: number) => void;
-    min: number;
-    max?: number;
-    readOnly: boolean;
-}
-
-const IntEdit: FunctionComponent<InitEditProps> = ({ value, onChange, min, max, readOnly }) => {
-    return (
-        <input
-            type="number"
-            value={value}
-            onChange={(e) => onChange(parseInt(e.target.value))}
-            min={min}
-            max={max}
-            readOnly={readOnly}
-            style={{ width: "4em" }}
-        />
-    )
-}
-
-type FloatEditProps = {
-    value: number;
-    onChange: (value: number) => void;
-    min: number;
-    max?: number;
-    readOnly: boolean;
-}
-
-const FloatEdit: FunctionComponent<FloatEditProps> = ({ value, onChange, min, max, readOnly }) => {
-    return (
-        <input
-            type="number"
-            value={value}
-            onChange={(e) => onChange(parseFloat(e.target.value))}
-            min={min}
-            max={max}
-            readOnly={readOnly}
-            style={{ width: "4em" }}
-        />
-    )
-}
-
-type IntOrUndefinedEditProps = {
+type NumberEditProps = {
     value: number | undefined;
     onChange: (value: number | undefined) => void;
     min: number;
+    max?: number;
     readOnly: boolean;
+    type: 'int' | 'float' | 'intOrUndefined';
 }
 
-const IntOrUndefinedEdit: FunctionComponent<IntOrUndefinedEditProps> = ({ value, onChange, min, readOnly }) => {
+const NumberEdit: FunctionComponent<NumberEditProps> = ({ value, onChange, min, max, readOnly, type }) => {
+    const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+        let newValue: number | undefined;
+        switch (type) {
+            case 'int':
+                newValue = parseInt(e.target.value);
+                break;
+            case 'float':
+                newValue = parseFloat(e.target.value);
+                break;
+            case 'intOrUndefined':
+                newValue = parseInt(e.target.value);
+                if (isNaN(newValue)) {
+                    newValue = undefined;
+                }
+                break;
+        }
+        onChange(newValue);
+    };
+
     return (
         <input
             type="number"
             value={value === undefined ? "" : value}
-            onChange={(e) => {
-                const value = parseInt(e.target.value);
-                onChange(isNaN(value) ? undefined : value)
-            }}
+            onChange={handleChange}
             min={min}
+            max={max}
             readOnly={readOnly}
             style={{ width: "4em" }}
         />
--- a/gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
+++ b/gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
@@ -102,6 +102,7 @@ const IntEdit: FunctionComponent<InitEditProps> = ({ value, onChange, min, max, 
             min={min}
             max={max}
             readOnly={readOnly}
+            style={{ width: "4em" }}
         />
     )
 }
@@ -123,6 +124,7 @@ const FloatEdit: FunctionComponent<FloatEditProps> = ({ value, onChange, min, ma
             min={min}
             max={max}
             readOnly={readOnly}
+            style={{ width: "4em" }}
         />
     )
 }
--- a/gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
+++ b/gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
@@ -77,7 +77,7 @@ const SamplingOptsPanel: FunctionComponent<SamplingOptsPanelProps> = ({ sampling
                         />
                     </Grid>
                 </Grid>
-                <Grid container item xs={12} spacing={sp2} title="Radius of the hypercube from which initial values for the model parameters are drawn (I think)">
+                <Grid container item xs={12} spacing={sp2} title="Radius of the hypercube from which initial values for the model parameters are drawn">
                     <Grid item xs={6}>
                         init radius
                     </Grid>
--- a/gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
+++ b/gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
@@ -76,6 +76,17 @@ const SamplingOptsPanel: FunctionComponent<SamplingOptsPanelProps> = ({ sampling
                             />
                         </td>
                     </tr>
+                    <tr>
+                        <td>seed</td>
+                        <td>
+                            <IntOrUndefinedEdit
+                                value={samplingOpts.seed}
+                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, seed: value })}
+                                min={0}
+                                readOnly={readOnly}
+                            />
+                        </td>
+                    </tr>
                 </tbody>
             </table>
             <div>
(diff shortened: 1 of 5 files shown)


gui/src/app/StanSampler/StanSampler.ts
```diff
@@ -58,7 +72,15 @@ class StanSampler {
         }
         this.#worker.postMessage({ purpose: Requests.Load, url: this.compiledUrl });
     }
-    sample(sampleConfig: Partial<SamplerParams>) {
+    sample(data: string | object, samplingOpts: SamplingOpts) {
+        const sampleConfig: SamplerParams = {
```
WardBrian, line 76: If you keep the mains consistent as you have thus far, this can be

```
        const sampleConfig: Partial<SamplerParams> = {
            data,
            ...samplingOpts
        }
```

Which I think also fixes the issue I'm having where it's trying to import `defaultSamplerParams` from tinystan which is private

magland: Yes I just fixed that, thanks.

magland:
@WardBrian So this addresses at least the basic opts you requested in #16 . LMK feedback on the controls

magland:
Does not include "initialization" that you mentioned... I wasn't sure how to do that.

WardBrian:
Yeah, that will require a new input area (maybe just tabs alongside the data?), I think.

Looking at it now, it's nice! Things seem to get a little squirrelly if you resize the window to be too narrow 

I think it's also worth adjusting the `refresh` parameter automatically. Here's something that seems like it produces reasonable values:

```ts

const calculateReasonableRefreshRate = (samplingOpts: SamplingOpts) => {
    const totalSamples = (samplingOpts.num_samples + samplingOpts.num_warmup) * samplingOpts.num_chains;

    const onePercent = Math.floor(totalSamples / 100);

    const nearestMultipleOfTen = Math.round(onePercent / 10) * 10;

    return Math.max(1, nearestMultipleOfTen);
}
```

This tries to target a number for `refresh` that is ~1% but is also "round" enough that it doesn't stand out

WardBrian:
It's also worth thinking about having the examples store default settings, if that's a reasonable request. That way we could have 'expensive' models like the SIR example default to something a lot smaller than 1000 warmup / 1000 sampling

magland:
Just added "seed" opt which can be null (by default) if left blank

magland:
Added the refresh as you suggested @WardBrian

magland:
> It's also worth thinking about having the examples store default settings, if that's a reasonable request. That way we could have 'expensive' models like the SIR example default to something a lot smaller than 1000 warmup / 1000 sampling

Yes I'm thinking about how to do that. It will be a different PR.

WardBrian:
After playing around with it for a bit I think `Math.max(10, nearestMultipleOfTen)` is a more reasonable value for `refresh` rather than `1`, which jumps around a lot.

I've also realized that when you're only running one chain, Stan leaves off the `Chain 1` from the start of the `Chain 1 Iteration 100 / 1000...` messages, which means the progress doesn't get reported properly. 

That should be an easy-ish fix. In the mean time, it revealed that the progress bar is still not properly reset before being displayed, as if you do a normal run and then a run with 1 chain it will show a broken progress bar the entire time

magland:
I'm not going to handle "initialization" input in this PR because that's a larger change. LMK if this looks okay to merge as is.

magland:
> After playing around with it for a bit I think `Math.max(10, nearestMultipleOfTen)` is a more reasonable value for `refresh` rather than `1`, which jumps around a lot.

I made this adjustment.

> I've also realized that when you're only running one chain, Stan leaves off the `Chain 1` from the start of the `Chain 1 Iteration 100 / 1000...` messages, which means the progress doesn't get reported properly.

I fixed that.

> That should be an easy-ish fix. In the mean time, it revealed that the progress bar is still not properly reset before being displayed, as if you do a normal run and then a run with 1 chain it will show a broken progress bar the entire time

I'll look into that now...

magland:
I can't reproduce the issue of the progress bar not being reset. Maybe it's not able to be noticed now that the num_chains=1 case is fixed.

WardBrian:
I think probably that eliminates it. I wasn't able to fully solve it before lunch but I suspect it had to do with the fact that `progress` can never be set back to `undefined` for a given sampler

magland:
Okay to merge then?

WardBrian:
I think so, not sure if @jsoules was looking at this

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+            onChange={(e) => {
+                const value = parseInt(e.target.value);
+                onChange(isNaN(value) ? undefined : value)
+            }}
+            min={min}
+            readOnly={readOnly}
+            style={{ width: "4em" }}
+        />
+    )
+}
```
jsoules, line 164: This is really one type and one component, with some very minor variation. We can leave it to refactor later, but it would be nice to do it now.

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+// ))
+//  Initializations. This is tricky, can either be 1 json object or a list of num_chains JSON objects.
+//  "Initialization radius" (parameters not given an initial value directly are drawn from uniform(-R, R) on the unconstrained scale) (default 2.0, can be [0,
+// ), but in practice limiting to say 10 in the UI is probably fine)
+//  Seed. Any uint32.
+
+
+const SamplingOptsPanel: FunctionComponent<SamplingOptsPanelProps> = ({ samplingOpts, setSamplingOpts }) => {
+    const num_chains = samplingOpts.num_chains;
+    const readOnly = !setSamplingOpts;
```
jsoules, line 26: It may be better style to explicitly compare with `undefined`

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const handleReset = useCallback(() => {
+        setSamplingOpts && setSamplingOpts(defaultSamplingOpts)
+    }, [setSamplingOpts])
+    return (
+        <div>
+            <table>
```
jsoules, line 32: Not fussed about it right now, but it'd probably be better not to use `table` for layout.

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+                                onChange={(value) => setSamplingOpts && setSamplingOpts({ ...samplingOpts, seed: value })}
+                                min={0}
+                                readOnly={readOnly}
+                            />
+                        </td>
+                    </tr>
+                </tbody>
+            </table>
+            <div>
+                <Hyperlink onClick={handleReset} color="gray">reset</Hyperlink>
```
jsoules, line 93: Maybe prefer a button here?

Also, this doesn't seem to be working for me when I run with the linear regression model?

The sample output div seems to be rendering on top of this link--so it's possible that I just can't click it because that div is in the way.

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+    const readOnly = !setSamplingOpts;
+    const handleReset = useCallback(() => {
+        setSamplingOpts && setSamplingOpts(defaultSamplingOpts)
+    }, [setSamplingOpts])
+    return (
+        <div>
+            <table>
+                <tbody>
+                    <tr>
+                        <td># chains</td>
```
jsoules, line 35: Future TODO: add tooltips explaining these parameters.

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -44,6 +47,18 @@ const HomePage: FunctionComponent<Props> = ({ width, height }) => {
         localStorage.setItem('data.json', dataFileContent)
     }, [dataFileContent])
 
+    const [samplingOptsContent, setSamplingOptsContent] = useState(initialSamplingOptsContent)
+    useEffect(() => {
+        localStorage.setItem('samplingOpts.json', samplingOptsContent)
```
jsoules, line 52: I assume setting this in local storage is to make it persistent between refreshes? I'm not sure that's the best behavior--it seems kind of weird that I can't get the defaults back by refreshing the page.

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -16,14 +18,15 @@ type Props = {
 const defaultStanContent = ''
 const defaultDataContent = ''
 const defaultMetaContent = '{"title": "Untitled"}'
+const defaultSamplingOptsContent = JSON.stringify(defaultSamplingOpts, null, 2)
```
jsoules, line 21: Is it possible to manually inspect this json? Just wondering about manually setting the indent spacing.

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -193,25 +216,34 @@ const LowerRightView: FunctionComponent<LowerRightViewProps> = ({ width, height,
             return undefined
         }
     }, [dataFileContent])
-    const runPanelHeight = 80
+    const samplingOptsPanelHeight = 140
+    const samplingOptsPanelWidth = Math.min(150, width / 2)
```
jsoules, line 220: This might be why I'm getting panel overlap from I think the SamplerOutputView--probably better to let the browser lay the page out rather than setting a fixed height.
Given the sizes at which these UI components are rendering, it would also not be unreasonable to have them horizontally between the data.json panel and the "run sampling" & "cancel" buttons.

Alternatively, we could have the entire sampler opts be a slide-in panel.

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -193,25 +216,34 @@ const LowerRightView: FunctionComponent<LowerRightViewProps> = ({ width, height,
...
 
     const {sampler} = useStanSampler(compiledMainJsUrl)
-
+    const {status: samplerStatus} = useSamplerStatus(sampler)
+    const isSampling = samplerStatus === 'sampling'
     return (
         <div style={{position: 'absolute', width, height}}>
-            <div style={{position: 'absolute', width, height: runPanelHeight}}>
+            <div style={{position: 'absolute', width: samplingOptsPanelWidth, height: samplingOptsPanelHeight}}>
```
jsoules, line 227: Future TODO: refactor hard-coded styles into CSS classes.

jsoules (commented):
Mix of suggestions and future TODOs that we should convert into issues once this is in.

Regarding the layout issue I was referring to, here's what I'm seeing:
image

The "reset" link (in the center) has the tabs of the sampling output display rendering on top of it, which may be making it impossible to click.

magland:
> Mix of suggestions and future TODOs that we should convert into issues once this is in.
> 
> Regarding the layout issue I was referring to, here's what I'm seeing: image
> 
> The "reset" link (in the center) has the tabs of the sampling output display rendering on top of it, which may be making it impossible to click.

Are you sure you have the latest version? I'm not seeing that overlap in my browser. Which browser are you using?

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
magland: Good idea. I have combined them.

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
magland: Done.

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
magland: I think a table might be okay here. But you are welcome to take a crack at a different approach.

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
magland: I have added tooltips.

gui/src/app/pages/HomePage/HomePage.tsx
magland: It's not visible to the user. I have removed the `null, 2` part

gui/src/app/pages/HomePage/HomePage.tsx
magland: Yeah, we can explore different layout options in a different PR.

gui/src/app/pages/HomePage/HomePage.tsx
magland: Yeah I here what you are saying. We can work on this in a separate issue.

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
WardBrian: I think eventually a `Grid` from materialUI would be nice, since it lets you do some responsive sizing relatively easily. 

We probably never expect someone to use this on a phone, but maybe an ipad is reasonable if we get our layout right

jsoules:
> > The "reset" link (in the center) has the tabs of the sampling output display rendering on top of it, which may be making it impossible to click.
> 
> Are you sure you have the latest version? I'm not seeing that overlap in my browser. Which browser are you using?

Just pulled again and the issue is still  present for me. This is Firefox 125.0.2, via snap. The issue isn't as obvious in Chrome, but the sample-reporting div--the outer wrapper in `LowerRightView`--is still laid out partially over the "reset" link.

By eye, it looks like Firefox wants to put more padding between the table rows by default than Chrome does, which may be why 140px is adequate in one browser but not the other.

In the short term, bumping the control panel height to 160px or something will (probably) fix it for my Firefox, but we will still likely see variations between browsers, monitor resolutions, and user font size choices, so the more permanent fix is probably to switch to a `Grid` like Brian suggested or find some other way to avoid the need for absolute positioning.

(Incidentally, making the comparison, I noticed that the adjustment arrows for the input boxes are always shown on Firefox, but in Chrome they only appear while the input box has focus. I don't propose anyone do anything with this information, I just thought it was interesting how much cross-browser variation there still is in 2024.)

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
jsoules: Thanks, that looks great.
One tweak--the "seed" tooltip says "Leave blank for a random seed"--we might want to say "Leave blank (not 0) for a random seed" or similar, since users may not realize that "0" is a specific seed value.

magland: done.

gui/src/app/SamplingOptsPanel/SamplingOptsPanel.tsx
jsoules: That's reasonable--we don't want it to draw too much attention for sure.

Would something like a Material UI React-Button in "Text button" variant be fit to purpose? (https://mui.com/material-ui/react-button/) I wouldn't swear to it as it's not my area of expertise, but I understand there may be accessibility reasons to prefer Button elements over raw links for form controls.

While it's out of scope for this PR, I think it'd be good to use MUI buttons over raw HTML buttons across the board; we could think about that and defer making any changes to this here in favor of doing so as part of a separate "use-MUI" PR later.

magland:
> > > The "reset" link (in the center) has the tabs of the sampling output display rendering on top of it, which may be making it impossible to click.
> > 
> > 
> > Are you sure you have the latest version? I'm not seeing that overlap in my browser. Which browser are you using?
> 
> Just pulled again and the issue is still present for me. This is Firefox 125.0.2, via snap. The issue isn't as obvious in Chrome, but the sample-reporting div--the outer wrapper in `LowerRightView`--is still laid out partially over the "reset" link.
> 
> By eye, it looks like Firefox wants to put more padding between the table rows by default than Chrome does, which may be why 140px is adequate in one browser but not the other.
> 
> In the short term, bumping the control panel height to 160px or something will (probably) fix it for my Firefox, but we will still likely see variations between browsers, monitor resolutions, and user font size choices, so the more permanent fix is probably to switch to a `Grid` like Brian suggested or find some other way to avoid the need for absolute positioning.
> 
> (Incidentally, making the comparison, I noticed that the adjustment arrows for the input boxes are always shown on Firefox, but in Chrome they only appear while the input box has focus. I don't propose anyone do anything with this information, I just thought it was interesting how much cross-browser variation there still is in 2024.)

Okay, I have switched from table to Grid. Let me know if that's working now.

jsoules:
> Okay, I have switched from table to Grid. Let me know if that's working now.

Thanks for making this change--but we do also need to remove (or change) the absolute positioning for the `LowerRightView` or that'll still get laid out on top.

It looks like there's a lot of absolute positioning going on for several layers of components--in `HomePage`, `SamplerOutputView`, and `TabWidget` at least. I'd suggest that for now we just increase the height from 140px to 160px, and open a separate issue to convert from absolute positioning to Flex or something.

magland:
> > Okay, I have switched from table to Grid. Let me know if that's working now.
> 
> Thanks for making this change--but we do also need to remove (or change) the absolute positioning for the `LowerRightView` or that'll still get laid out on top.

Have you tested that? I was hoping Grid would fix that on your browser. If not, I'll increase to 160.

I agree absolute positioning has downsides... but it will take a lot to convert everything to flex, and then we'll lose some of the control.

jsoules:
> > Thanks for making this change--but we do also need to remove (or change) the absolute positioning for the `LowerRightView` or that'll still get laid out on top.
> 
> Have you tested that? I was hoping Grid would fix that on your browser. If not, I'll increase to 160.

I did check it, yeah. The `absolute` layout instruction will trump any other page content. We *could* mess with the z-index, but then you'd just have one of the tabs be un-clickable instead of the reset button being un-clickable.

> I agree absolute positioning has downsides... but it will take a lot to convert everything to flex, and then we'll lose some of the control.

Agreed--that's why I think we should table it for now and follow up in a separate issue. There may be some role for absolute positioning to play within the component--e.g. keeping the tab bar tight against the content pane--while deferring a lot of the other page flow to the browser, if one of the parents has relative positioning; but getting this right is always very finicky and involved.

magland:
Okay @jsoules I've increased that hard-coded value to 160 pixels. I'm going to go ahead and merge this now... but feel free to continue to follow up with separate issues.

PR #29, opened 2024-05-30
Simplify docker image with newer tinystan

WardBrian:
TinyStan has the `.js` target now, so we don't need the `make/js` file.


jsoules (approved):
Changes look good, we've discussed already, and everything seems to still work with the new image, so I say push it

PR #30, opened 2024-05-31
Run build in current job folder

WardBrian:
This PR makes some simplifications to the build setup that lead to slightly more consistent behavior.

The new flow of information is:

1. A request to build a model comes in
2. We check to see if it has already been built and cached in the $MODEL_HASH folder
   a. If so, we return*
3. Assuming it was not a cache hit, we build the model _in the job folder_. This means a job always has something to do, removing the case before where it would be told that compilation was already running and fail
4. Copy the output to the $MODEL_HASH folder*

The * above indicate places where we need to take or wait on a lock to make sure we aren't viewing incomplete copies on the filesystem, but in general this locking is simpler (and the critical section much smaller) than before.

This allowed us to remove some code that was tracking statuses. I also moved the exception registration into a more data-driven style that doesn't have as many repeated functions in the main.py file.

### DIFF
--- a/docker/stan-wasm-server/logic/compilation.py
+++ b/docker/stan-wasm-server/logic/compilation.py
@@ -25,18 +25,6 @@ def make_canonical_model_dir(src_file: Path):
     return model_dir.absolute()
 
 
-# TODO: Remove this and run directly from the canonical model directory
-def copy_compilation_outputs(*, model_dir: Path, job_dir: Path):
-    todos = []
-    for f in COMPILATION_OUTPUTS:
-        src = model_dir / f
-        dest = job_dir / f
-        if not src.exists():
-            raise FileNotFoundError(f"File {src} does not exist.")
-        todos.append((src, dest))
-    for (src, dest) in todos:
-        copy2(src, dest)
-
 async def compile_and_cache(*, job_id: str, model_dir: Path, tinystan_dir: Path):
 
     # if there's a cache hit, make sure any copying is already complete,
@@ -56,7 +44,8 @@ async def compile_and_cache(*, job_id: str, model_dir: Path, tinystan_dir: Path)
             for file in COMPILATION_OUTPUTS:
                 copy2(job_dir / file, model_dir / file)
         # otherwise, just wait for the other thread's version to be
-        # copied. We wasted some time, but that's ultimately okay
+        # copied. We do not need to copy, because their version will
+        # be equivalent; we wasted some time, but that's ultimately okay
         else:
             wait_until_free(model_dir)
 
--- a/docker/stan-wasm-server/logic/compilation.py
+++ b/docker/stan-wasm-server/logic/compilation.py
@@ -1,16 +1,15 @@
 import asyncio
 from hashlib import sha1
-from shutil import copy2, rmtree
+from shutil import copy2
 from pathlib import Path
 
-from .definitions import CompilationStatus
-from .locking import get_nonce, acquire_compilation_lock, release_compilation_lock
+from .locking import compilation_output_lock, wait_until_free
 from .exceptions import StanPlaygroundCompilationException, StanPlaygroundCompilationTimeoutException
-
+from .compilation_job_mgmt import get_job_source_file, get_compilation_job_dir
+from .file_validation.compilation_files import COMPILATION_OUTPUTS
 
 COMPILED_MODELS_BASE_PATH = Path('/compiled_models')
 COMPILATION_TIMEOUT = 60 * 5
-COMPILATION_OUTPUTS = ['main.js', 'main.wasm']
 
 
 def _compute_stan_program_hash(program_file: Path):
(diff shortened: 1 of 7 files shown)


WardBrian:
To test this, I inserted a loop after the copy in the critical section that just `time.slept` for ~30 seconds, printing as it went, and a print statement in the `wait_for_lock` function, and confirmed that these were doing what we expect.

docker/stan-wasm-server/logic/compilation.py
```diff
@@ -38,60 +37,46 @@ def copy_compilation_outputs(*, model_dir: Path, job_dir: Path):
     for (src, dest) in todos:
         copy2(src, dest)
```
jsoules, line 38: Is this section still needed? The note says to remove it, but not sure where we landed.

docker/stan-wasm-server/logic/compilation.py
```diff
@@ -38,60 +37,46 @@ def copy_compilation_outputs(*, model_dir: Path, job_dir: Path):
...
 
+async def compile_and_cache(*, job_id: str, model_dir: Path, tinystan_dir: Path):
+
+    # if there's a cache hit, make sure any copying is already complete,
+    # then return
+    if all((model_dir / x).exists() for x in COMPILATION_OUTPUTS):
+        wait_until_free(model_dir)
+        return
```
jsoules, line 46: Are we doing the "check for all compilation outputs" thing elsewhere? (e.g. the possibly-removed `copy_compilation_outputs` function above)

If so, could potentially standardize the file check.

docker/stan-wasm-server/logic/compilation.py
```diff
@@ -38,60 +37,46 @@ def copy_compilation_outputs(*, model_dir: Path, job_dir: Path):
...
+    with compilation_output_lock(model_dir) as exclusive:
+        # if we succeed in getting the lock, do the copy
+        if exclusive:
+            job_dir = get_compilation_job_dir(job_id)
+            for file in COMPILATION_OUTPUTS:
+                copy2(job_dir / file, model_dir / file)
+        # otherwise, just wait for the other thread's version to be
+        # copied. We wasted some time, but that's ultimately okay
+        else:
+            wait_until_free(model_dir)
```
jsoules, line 61: Might want to explicitly note that we intentionally do not copy in this case (because somebody else succeeded at compiling something semantically equivalent)

docker/stan-wasm-server/logic/locking.py
```diff
@@ -1,37 +1,36 @@
...
+        acquired = False
 
-def release_compilation_lock(model_dir: Path, nonce: str):
-    lockfile = _get_compilation_lockfile_name(model_dir)
     try:
-        contents = lockfile.read_text()
-        if contents == nonce:
+        yield acquired
+    finally:
+        if acquired:
```
jsoules, line 33: Maybe feels better to do this as one try-catch instead of two? Also the `finally` is conditioned on successfully completing the `try` block. Maybe:

```python
lockfile = _get_compilation_lockfile_name(model_dir)

try:
    with lockfile.open(mode='x') as file:
        file.write("locked")
        yield True
        lockfile.unlink(missing_ok=True)
    except:
        yield False
```

Hm, what happens if there's an exception elsewhere after we `yield`--would that result in failure to unlock? If so we'd need to put the `finally` back.

jsoules (commented):
We mostly went over this already, and I like it. A couple things jumped out while rereading, but I think we're good.

magland:
> This PR makes some simplifications to the build setup that lead to slightly more consistent behavior.
> 
> The new flow of information is:
> 
> 1. A request to build a model comes in
> 2. We check to see if it has already been built and cached in the $MODEL_HASH folder
>    a. If so, we return*
> 3. Assuming it was not a cache hit, we build the model _in the job folder_. This means a job always has something to do, removing the case before where it would be told that compilation was already running and fail
> 4. Copy the output to the $MODEL_HASH folder*
> 
> The * above indicate places where we need to take or wait on a lock to make sure we aren't viewing incomplete copies on the filesystem, but in general this locking is simpler (and the critical section much smaller) than before.
> 
> This allowed us to remove some code that was tracking statuses. I also moved the exception registration into a more data-driven style that doesn't have as many repeated functions in the main.py file.

This makes a lot of sense to me.

Maybe in the future we could have the client side simply query for that existing hash url and not create a job at all if found.

docker/stan-wasm-server/logic/compilation.py
WardBrian: Yeah, dead code

docker/stan-wasm-server/logic/compilation.py
WardBrian: Not currently

docker/stan-wasm-server/logic/locking.py
WardBrian: We need two `try`s somewhere - they can be nested or in sequence, but the unlock needs to happen in a `finally` in case anything throws during the `yield`

docker/stan-wasm-server/logic/compilation.py
WardBrian: I realized another condition which could lead to an error, so this is now pulled out.

The issue was another sequencing one:

```
THREAD A starts compiling
THREAD B starts compiling
THREAD A finishes
THREAD A acquires the lock
THREAD A copies the files
THREAD A releases the lock

THREAD B finishes
THREAD B acquires the lock
THREAD B begins copying, WHILE THREAD A requests to download the files

THREAD A reads partially completed copies
```
This is fixed by having THREAD B check if the files exist before copying them, and I've tried to explain it in a comment in the file

docker/stan-wasm-server/logic/locking.py
WardBrian: Made this into a nested try, which looks cleaner IMO

gui/src/app/compileStanProgram/compileStanProgram.ts
```diff
@@ -45,7 +45,8 @@ const compileStanProgram = async (stanWasmServerUrl: string, stanProgram: string
             }
         });
         if (!c.ok) {
-            onStatus(`failed to run job: ${c.statusText}`);
+            let j = await c.json();
+            onStatus(`failed to run job: ${j.message ?? c.statusText}`);
```
jsoules, line 49: Sorry, probably best to do `j?.message ?? c.statusText` where the `?.` is a null propagation operator (not sure if that's the official name)--otherwise there's a potential error in the case where `j` is undefined, and thus has no members, so `j.message` cannot resolve.

jsoules (commented):
Sorry, 1 more change. Otherwise looks good.

docker/stan-wasm-server/logic/compilation.py
jsoules: Yep, want to check for existence both before and after getting the lock. Good call.

jsoules (approved):
Yeah, this looks good, as far as I can see. Ship it!

PR #20, opened 2024-05-16
[Obsolete] Share button

magland:
Add a "Share" button on the left. This uploads the main.stan, data.json, and sampling opts to the wasm compiler server. Those are stored in a content-addressable storage system - so that the same stan/data.json/etc only gets stored once. A url is then provided that is of the form

[https://stan-playground.vercel.app/?stan=c8799ed2f8bd587833f979b9d46b81c0fa6ee4b2&data=60b0eb9f7622ccddda78b37d38ad3b2912db8047&sopts=19462ddf12967beccf1f9fad14fc07776f0fcd76&title=Linear regression](https://stan-playground.vercel.app/?stan=c8799ed2f8bd587833f979b9d46b81c0fa6ee4b2&data=60b0eb9f7622ccddda78b37d38ad3b2912db8047&sopts=19462ddf12967beccf1f9fad14fc07776f0fcd76&title=Linear%20regression)

You can see the `stan`, `data`, and `sopts` query parameters are sha1 hashes of the contents which are stored on the wasm server. The `title` query parameter is also provided.

To persist the data on the server, you can do

```
docker run -p 8083:8080 -v $HOME/stan-playground-blobs:/blobs -e BLOBS_DIR=/blobs -it magland/stan-wasm-server:latest
```

This is spelled out in the new README

Issue #11, opened 2024-05-15
Page boundary issues


WardBrian:
A few things seem eager to escape their viewport/not wrap/scroll as you might expect:

- The sample output area (including the table in #10) - falls off the end of the page, no option to scroll
- The JSON editor doesn't have text wrapping, and if I enable it with the parameter to the Editor component it seems to wrap at the wrong place (e.g. part is still off the right hand side of the screen)

WardBrian:
Note: might be a duplicate of #24

WardBrian:
This was resolved by #135

PR #31, opened 2024-05-31
[obsolete] save to cloud or gist via jpfiddle infrastructure

magland:
This reorganizes the persisting of analyses by piggy-backing on [jpfiddle](https://github.com/magland/jpfiddle).

With jpfiddle you can save directories of code (text files) in either the cloud (in a cloud bucket) or as github gists. In order to avoid having another server-side component, and reduce the duplication of code and infrastructure, this PR has SP utilize the jpfiddle database and API. (An added benefit is that you can open any SP analysis as a jpfiddle fiddle and browse and edit it using jupyterhub.)

The left panel has been replaced, and is equal to the left panel of jpfiddle, with some different configuration options. This includes basic info about the analysis/fiddle, list of recent analysis, buttons to save to the cloud, buttons to save as gist, and a list of examples. The two current examples are stored in gists.

To get an idea of how everything looks and fits together, here's one example as a SP analysis: (I've deployed to production -- we're at early stage so I feel okay doing this)
https://stan-playground.vercel.app/?p=/&f=https://gist.github.com/magland/d5bc7d37561539dca14256061c829cc9&t=SP%3A%20Disease%20transmission

The same viewed using jpfiddle
https://jpfiddle.vercel.app/?f=https://gist.github.com/magland/d5bc7d37561539dca14256061c829cc9&t=SP%3A%20Disease%20transmission

And viewed as a gist
https://gist.github.com/magland/d5bc7d37561539dca14256061c829cc9

Saving as a gist is good for long-term persistence. But if you want to quickly save your work, you can "save to cloud", which will store it temporarily in a cloud bucket. In order to do this, you need to know a secret passcode, which I can tell you.

Your edits for each analysis are saved in the browser storage, and the left panel shows which files have been changed compared with the cloud/gist version. You can discard those changes by using the "reset from cloud version" button.


magland:
@WardBrian @jsoules 

I put you as reviewer, but I don't expect you to be too thorough looking through the frontend code, since there are a lot of complex changes. Happy to answer any questions.

WardBrian:
This replaces #20, right?

Looking at it now, it seems like the UI changes have lead to even worse alignment/overlapping issues than what I've previously seen in #11. See the green checkmark's location here:

image

WardBrian:
A couple other general thoughts:
- is a Github PAT really the best authentication method we have? Is it possible to get a "log in with Github" page?
   - if PATs are our best choice, we should explicitly state which of the [fine-grained permissions](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#fine-grained-personal-access-tokens) it needs
- Do you think that some of the shared infrastructure will end up in an npm package?

WardBrian:
Did we lose the "Reset local storage" button?

magland:
> A couple other general thoughts:
> 
> * is a Github PAT really the best authentication method we have? Is it possible to get a "log in with Github" page?

I agree the PAT method is not ideal. I did try creating a proper github app, with log in window etc. But there didn't seem to be the option of limiting the scope of the permissions to gists. Even if that were possible, I as a user would not want to trust another entity with permission to edit/delete/add all gists. The PAT is also a matter of trust, but it's open source and one can see that the code does not store the token anywhere. At this point I don't think there is a better way unfortunately.

>   
>   * if PATs are our best choice, we should explicitly state which of the [fine-grained permissions](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#fine-grained-personal-access-tokens) it needs

I agree. This will require making a different kind of modal dialog that can display this information. Will require more work compared to just a system popup.

> * Do you think that some of the shared infrastructure will end up in an npm package?

Eventually yes I think so. Too much in flux right now though.

magland:
@WardBrian Regarding GitHub PAT. I'm going to take a crack at making this more user-friendly, and experiment with allowing the user to save the token in the browser storage for convenience, while giving them a warning about the potential security vulnerability. It is possible to limit the scope of the token to gists only, which I think helps a lot.

As I mentioned, I don't think it's possible to do a proper login as a github app while restricting to gist-only scope. I'll look into that further though.

magland:
Replaced by a different series of PRs.

#40 #41 #42 #43

PR #34, opened 2024-06-03
Backend: use Pydantic Settings object for config

### LINKED ISSUE #33: Backend: App settings
See [Settings and Environment Variables](https://fastapi.tiangolo.com/advanced/settings/#settings-and-environment-variables) in the FastAPI docs

We currently have the following items either hardcoded or using environment variables:

- `SWS_PASSCODE`
- `BASE_JOB_DIR`
- `COMPILED_MODELS_BASE_PATH`
- `COMPILATION_TIMEOUT`
- `TINYSTAN_DIR`

Additionally, it would be nice to have a global `DEBUG` setting, and possibly gate the idea of canonicalizing models for cache purposes behind a feature flag, since there are known tradeoffs to doing so. 

Each of these being runtime configurable would also help with testing, see the FastAPI note on the subject: https://fastapi.tiangolo.com/advanced/settings/#settings-and-testing

WardBrian:
Closes #33.

This adds a `config.py` which stores the global settings for the project. 

A few functions needed to be rethought/changed to take in these settings rather than depend on globals, which is the reason for most of the changes outside `main.py`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -31,6 +31,10 @@ RUN mkdir oneTBB/build && cd oneTBB/build && \
     cmake --build . -j$(nproc) && \
     cmake --install .
 
+# Install Python dependencies
+RUN pip install fastapi[all]
+RUN pip install uvicorn
+
 # Clone the TinyStan repository and checkout a specific commit
 RUN git clone https://github.com/WardBrian/tinystan.git && \
     cd tinystan && \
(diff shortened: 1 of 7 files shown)


jsoules (commented):
Code changes all make sense and compilation succeeds with local server running this code. I think we're good here, but if we're feeling less confident, it might be good to actually test that we're actually picking up environmental config for `job_dir`, `build_model_dir`, `compilation_timeout`. If that's a feature we actually care about.

WardBrian:
> we're actually picking up environmental config for `job_dir`, `build_model_dir`, `compilation_timeout`. If that's a feature we actually care about.

I checked that locally, but it should eventually get done, yes. Though to a certain extent that's testing Pydantic, no?

PR #37, opened 2024-06-05
Add test framework with coverage & minimal example.

jsoules:
This PR adds a framework for writing tests against the Typescript project. Changes are in the `gui` directory.

- `.gitignore` the coverage reports
- `package.json`: add `vitest`, `vitest/coverage-v8` packages and update defined scripts to include test commands (`test` to run once, `test-watch` to run in interactive/watch mode and rerun with file changes)
- `tsconfig.json`: updating known paths for code
- `vite.config.ts`: some minor test configuration, for mock management and coverage inclusions/exclusions.
- Example of a test in `vitest` format. (I have examples for interacting with `react` elsewhere but wanted to keep it simple for now)
- I upgraded to Yarn 2 a few months back, which probably explains the massive diff for the `yarn.lock` file. We may need to back out the changes to that file if they break things for others.

There's also an incidental change to `gui/src/app/compileStanProgram/compileStanProgram.ts` where linting was complaining about something that could be `const` instead of `let`. It's a drive-by, I probably shouldn't have changed it...


magland:
Cool! Looks good to me.

jsoules:
Per conversation w/ @WardBrian , changed the example; this also demonstrates testing an exception & exporting "private" code for test-only.

Note that this method of handling private exports does confuse the code coverage module--it can't tell that the private function is actually tested, only the object faking the export. So coverage reports will be misleading.

WardBrian:
> Note that this method of handling private exports does confuse the code coverage module--it can't tell that the private function is actually tested, only the object faking the export. So coverage reports will be misleading.

Oh that's too bad. We can keep an eye on how often we actually want to do this kind of thing.
I'm still working on the side on re-organizing the tinystan code, which ends up exporting this function anyway, so we might not really need that trick much if at all.

jsoules:
> Oh that's too bad. We can keep an eye on how often we actually want to do this kind of thing. I'm still working on the side on re-organizing the tinystan code, which ends up exporting this function anyway, so we might not really need that trick much if at all.

Actually `coverage-gutters` is making a liar of me--I toggled it off and on and it is labeled, and looking at the actual coverage report it's covered. So false alarm, I think!

WardBrian (approved):

PR #38, opened 2024-06-05
Backend: add basic logging

WardBrian:
This adds some usages of Python's built-in `logging` module. 

This adds logs for when
- compilation starts
- compilation ends
- we're copying files
- we hit the cache

I set this up to use the same format as `uvicorn`'s pre-existing logs

I also added some debug-level logs for the locking mechanism

### DIFF
--- a/backend/stan-wasm-server/src/app/config.py
+++ b/backend/stan-wasm-server/src/app/config.py
@@ -1,9 +1,11 @@
 import logging
 from functools import lru_cache
 from pathlib import Path
+from typing import Annotated, Literal
 
 from pydantic import (
     AliasChoices,
+    BeforeValidator,
     DirectoryPath,
     Field,
     PositiveInt,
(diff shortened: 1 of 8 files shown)


backend/stan-wasm-server/src/app/config.py
```diff
@@ -22,6 +23,15 @@ class StanWasmServerSettings(BaseSettings):
...
         validation_alias=AliasChoices("tinystan", "tinystan_dir")
     )
+    log_level: str = "INFO"
+
+    @field_validator("log_level")
+    @classmethod
+    def log_level_is_valid(cls, v: str) -> int:
+        v = v.upper()
+        if v not in ("DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"):
+            raise ValueError(f"Invalid log level: {v}")
```
jsoules, line 33: This is a little magic--validating the (string) envvar value, then looking up against the module for a (hopefully extant) numeric code corresponding to that value, then preserving that in the settings object (despite the field being annotated as `str`). Might want to leave a note for this :)

jsoules (approved):
Code looks good, confirmed logging is visible in the console.

backend/stan-wasm-server/src/app/config.py
WardBrian: I re-thought this, give it a fresh look?

jsoules: I think this is a lot clearer--nice.

PR #43, opened 2024-06-06
Track local changes

magland:
This builds on #42 . It tracks local edits (although these are still not saved to browser storage) and displays which files have been modified in the left panel.


magland:
Note: This is not about the UI - we may or may not want to display this information for the user - but being able to track changes is important for next steps for sharing to cloud/gist etc.

magland:
Closing because we are going in a different direction.

PR #44, opened 2024-06-07
Hide syntax error window by default

magland:
@WardBrian @jsoules 

Changes
* The syntax error window attached to the Stan file editor is hidden by default
* If there is a syntax error, an icon/button appears in the toolbar
* If user clicks on the icon, the syntax window opens, and stays open while the user types
* The syntax window contains a close button
* If the stan file editor has no text, then there is a different window that opens below instructing the user to open an example to the left.

I created this on top of #40, the stripped down version that abstracts the data model, hoping that can be a base point until we decide how to proceed on that front.

Before button click:
image

After button click:
image


magland:
Replaced by #51 which is based on main branch. Closing.

PR #45, opened 2024-06-07
[NEED REBASE ON MAIN] Adjust main layout

magland:
As we had discussed...

* Move the data editor window to the lower left
* Sampling controls and sampling output now has more space on the right side

Built on #44 because hiding syntax window by default is important to give more space to the Stan editor

image


magland:
Replaced by #102

PR #32, opened 2024-05-31
[obsolete] Update sampling output window

magland:
This is a redo of https://github.com/flatironinstitute/stan-playground/pull/21

I took all those changes and put them on top of the new chain of changes - since the file sharing has been revamped.

See the discussion in the other PR. I think there are still some things to check/improve, but I think it would be great to merge this for now.

PR #46, opened 2024-06-07
develop sampling output view

magland:
This applies the changes of #32 onto the new base #40 (again, hoping that can be a starting point)

I realize there are some questions about the correctness of the items.

What I would suggest so that we can move forward until we can get the proper tests in place

* Hide the controversial columns (N_eff, N_eff/s, R_hat) for the time being
* Verify the other columns - that should be pretty straightforward
* Put them back in once we have proper tests

That way we can benefit from most of this improved view while we do further dev.

What do you think @WardBrian @jsoules ?


magland:
> Two thoughts:
> 
> 1. Does this actually depend on [implement general data model - agnostic starting point for further development #40](https://github.com/flatironinstitute/stan-playground/pull/40), or could it be done independently? It seems to me like it could, but I might be missing something

Technically it could, but I'm really hoping we can merge 40 first - it doesn't add features, it simplifies and removes things so that we can be at a base starting point for proposed development. Looking at the files changed, it's true I wouldn't anticipate any conflicts... but I hope we can just merge 40 and then it's not an issue.

> 2. Rather than merge ess code we don't want to use, could we just pull the plots tabs into their own PR and not do anything to the summary tab yet?

I can move the ess stuff into a separate PR... But shall we keep the table with the non-controversial columns? That way the ess PR can be pure to that.

WardBrian:
I think I spotted a meaningful disagreement between Stan's summary and this table in every column. Some may just be up to different implementation choices for e.g. quantile computation, but I'd like to look into them more.

As for depending on #40, I think it is always better to not have interdependence when it isn't needed. I'd be happy to merge a plots-only based-on-main version of this today, but if it is dependent on another PR that will naturally take more time

magland:
> I think I spotted a meaningful disagreement between Stan's summary and this table in every column. Some may just be up to different implementation choices for e.g. quantile computation, but I'd like to look into them more.
> 

Okay sounds good... I'll separate all that out.

> As for depending on #40, I think it is always better to not have interdependence when it isn't needed. I'd be happy to merge a plots-only based-on-main version of this today, but if it is dependent on another PR that will naturally take more time

I think it's okay to wait. If we end up not merging #40, then I'll make the adjustments at that point.

PR #49, opened 2024-06-07
Refactor: split up tinystan/index.ts, add tests

WardBrian:
This breaks up `app/tinystan/index.ts` into several files:

- `types.ts` -- public types, like the arguments/return types to functions, and 'private'* types, like the WasmModule
- `constants.ts` -- default values of algorithmic arguments, names of HMC output variables, etc
- `util.ts` -- the `string_safe_jsonify` function, now called `prepareStanJSON` with a better type
- `model.ts` -- just the class definition of StanModel
- `index.ts` -- just re-exports relevant items from the above.

`*` - I settled on a perhaps-too-cute trick for the private/internal types, see comment in the file

And adds testing of the StanModel (currently at 100% coverage!)

This is based around a regrettably complicated mock object for the WasmModule type. It tests most of the "hard parts", namely:

- memory management is checked (e.g., each `malloc` is paired with exactly one `free`)
- output objects are in the expected shape/layout (column major vs row major)

Integration testing would require checking in a built model, which I will probably do once this has been factored out into its own repo.


gui/test/app/tinystan/mocking/WasmModule.ts
```diff
@@ -0,0 +1,82 @@
...
+// relative to each other
+const fakeHeap = new Float64Array(16 * 1024 * 1024);
+for (let i = 0; i < fakeHeap.length; i++) {
+  fakeHeap[i] = i;
+}
+
+export const mockModule = (p: Partial<ModuleSettings>) => {
+  const {
+    returnCode,
+    numFreeParams: numParams,
```
jsoules, line 33: Note the automatic aliasing VSCode "helpfully" did for you here (unless that was intentional)

(VSCode likes to Help(tm) when you use the F2/context-aware rename for object members that get unpacked)

gui/test/app/tinystan/model.test.ts
```diff
@@ -0,0 +1,427 @@
...
+        draws.forEach((d) =>
+          expect(d.length).toEqual(num_chains * (num_samples + num_warmup)),
+        );
+
+        expectColumnMajor(
+          draws,
+          N_PARAMS,
+          num_chains * (num_samples + num_warmup),
+        );
+      }
```
jsoules, line 167: It's not terribly important, but you could probably do this as a loop over `[true, false]` for `save_warmup`

gui/test/app/tinystan/model.test.ts
```diff
@@ -0,0 +1,427 @@
...
+    });
+
+    test("erroring call behavior", async () => {
+      const { mockedModule, model } = await getMockedModel({ returnCode: 1 });
+      expect(() => model.pathfinder({})).toThrow(/Error inside WASM/);
+
+      expect(mockedModule._tinystan_pathfinder).toHaveBeenCalledTimes(1);
+    });
+
+    test("empty model errors", async () => {
```
jsoules, line 257: Same note re: name ambiguity

gui/test/app/tinystan/model.test.ts
```diff
@@ -0,0 +1,427 @@
...
+      const { mockedModule, model } = await getMockedModel({
+        numFreeParams: 0,
+      });
+      expect(() => model.pathfinder({})).toThrow(/no parameters/);
+
+      expect(mockedModule._tinystan_pathfinder).toHaveBeenCalledTimes(0);
+    });
+
+    test("model failing to allocate errors", async () => {
```
jsoules, line 266: Same note re: name ambiguity

gui/test/app/tinystan/model.test.ts
```diff
@@ -0,0 +1,427 @@
...
+        1,
+        dataStr,
+        expect.any(Number),
+        expect.any(Number),
+      );
+
+      expect(mockedModule._tinystan_pathfinder).toHaveBeenCalledTimes(2);
+    });
+
+    test("output is correct shape and layout", async () => {
```
jsoules, line 303: Also looks like this could be a loop, if you don't mind some conditionals in the loop.

gui/test/app/tinystan/mocking/WasmModule.ts
```diff
@@ -0,0 +1,82 @@
...
+  returnCode: number;
+  paramNames: string;
+  numFreeParams: number;
+  modelFails?: boolean;
+};
+
+const defaultModuleSettings: ModuleSettings = {
+  returnCode: 0,
+  paramNames: "foo,bar,baz",
+  numFreeParams: 3,
```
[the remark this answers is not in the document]
jsoules: I guess you do use it where you specifically know that you're checking just the count and not the values, but that feels a little cheaty

jsoules (approved):
This seems good.

I have suggested a couple places where you could clarify wording for those who come after, and a few instances where you might consider reducing repetitive code a bit, but otherwise this is great.

gui/test/app/tinystan/mocking/WasmModule.ts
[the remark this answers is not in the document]
WardBrian: This actually is different! Consider the following Stan model:

```stan
parameters {

}
transformed parameters {
   real foo = 3.14;
}
```

The real tinystan wasm will report param names is `"foo"`, but the number of free parameters is `0`

gui/test/app/tinystan/mocking/WasmModule.ts
```diff
@@ -0,0 +1,82 @@
...
+    numFreeParams: numParams,
+    paramNames,
+    modelFails,
+  } = {
+    ...defaultModuleSettings,
+    ...p,
+  };
+
+  // malloc returns unique number - useful for checking leaks
+  // 8 aligned to mimic "real" double*
```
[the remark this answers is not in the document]
WardBrian: The code in the tested object actually does divide these numbers by 8 to calculate indices into the heap array

gui/test/app/tinystan/mocking/WasmModule.ts
WardBrian: Yeah this is the kind of "deeper" refactoring I was trying to ask you if I could have vscode do...

gui/test/app/tinystan/model.test.ts
```diff
@@ -0,0 +1,427 @@
...
+    expect(v).toEqual("123.123.123");
+    expect(mockedModule._tinystan_stan_version).toHaveBeenCalledTimes(1);
+  });
+
+  describe("sample function", () => {
+    test("null call behavior", async () => {
+      const { mockedModule, model } = await getMockedModel({});
+      model.sample({});
+
+      expect(mockedModule._tinystan_sample).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
WardBrian: Because all the tests check that no memory leaks occur, that is probably the primary thing this is actually asserting

gui/test/app/tinystan/model.test.ts
```diff
@@ -0,0 +1,427 @@
+import { describe, expect, test } from "vitest";
+import { getMockedModel } from "./mocking/getMockedModel";
+import { HMCMetric } from "../../../src/app/tinystan/types";
+import {
+  HMC_SAMPLER_VARIABLES,
+  PATHFINDER_VARIABLES,
+} from "../../../src/app/tinystan/constants";
```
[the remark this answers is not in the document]
WardBrian: leaving this out only because I'm going to move all this to its own project next

jsoules (approved):
Looks great.

PR #52, opened 2024-06-11
Switch to using tinystan from npm

WardBrian:
Follow on to #49, I've moved the tinystan code out and into its own repo, and uploaded to npm: https://www.npmjs.com/package/tinystan

Unfortunately this also removes the only examples of tests we have in the repo, working on a few new ones to add back


magland:
Nice!

magland (approved):

PR #65, opened 2024-06-13
Load data from URL query

magland:
This one implements part of the URL query scheme we decided on yesterday.

Builds on #40

Here's an example URL

http://127.0.0.1:3000/?stan=https://gist.githubusercontent.com/magland/da3d4143276827609ec9317bb3db8b04/raw/cf48846a36de0559f96a9e4bcb5a5dcbe46383be/main.stan&data=https://gist.githubusercontent.com/magland/da3d4143276827609ec9317bb3db8b04/raw/cf48846a36de0559f96a9e4bcb5a5dcbe46383be/data.json&num_chains=6&num_warmup=1000&num_samples=1000&init_radius=2&seed=6&title=test_title

It specifies stan, data, num_chains, num_warmups, num_samples, init_radius, seed, and title

All the allowed query parameters are: `stan, data, sampling_opts, num_chains, num_warmup, num_samples, init_radius, seed, title`

You can't specify both sampling_opts and the inline opts at the same time.

After the initial load, if any of the content of the project changes, the query string on the URL disappears (as we had decided). But if you undo the changes within the same session, the query reappears. I find that convenient.

If the project has been edited (and the query string disappears) the changes are saved to browser storage.

I have not yet implemented loading entire project from single gist URI. I think that should be a separate PR along with saving to gist.


magland:
Replaced by #84

Issue #12, opened 2024-05-15
Debounce calls to the javascript Stan compiler


WardBrian:
This is actually surprisingly fast even for the [jm.stan model](https://github.com/stan-dev/rstan/files/12554818/jm.txt), but freezing the UI to run the compiler does make typing feel unnatural after a certain size of model

WardBrian:
I don't really think this is necessary with stanc in a worker -- jm.stan is pretty snappy still, and nobody will be using this for something _bigger_ than ~3k lines

PR #66, opened 2024-06-13
Move stanc.js to a web worker

WardBrian:
This PR has a few contributions:

- we now vendor `stanc.js` at a fixed version, rather than getting the `nightly` build. This is currently 2.35.0, which is also what version of Stan our current tinystan is using in the backend. 
- rather than loading in a `script` tag, stanc.js is loaded in a worker. This gives us some degree of sandboxing because the worker thread's globals are separate from the rest of our app. 
- the state logic around this worker and around the errors from stanc was extracted out into a `useStanc` hook
- this hook is tested 

It is probably still worth it to debounce calls to stanc (#12), but this PR technically does resolve the issue with the UI slowing down when editing a large model.

To be honest, preparing this PR took much longer than it was probably worth, mainly on the final bullet point. There was a lot of pain associated with the fact that `vitest` is not really a browser, but in the end it seems to be working.


magland:
Very cool.

This all looks great to me.

(I tested it using "review in codespaces" and it was really easy... just `yarn install` and `yarn dev`. I recommend to test all the PRs this way because it does not require checking out locally)

WardBrian:
> "review in codespaces" and it was really easy

Wow, that really does feel like magic. The fact that it actually redirects localhost:3000 to another address so you can actually open it is very cool.

It looks like each user gets 60 hours/month if you are just on a standard free github plan. Pretty nice!

magland:
I confirmed that this branch does not conflict with #40

WardBrian:
It probably conflicts with #44, but I think it's fairly trivial to resolve no matter which one got merged first

magland:
This conflicts sort of heavily with #51

It's manageable but will require some focus to get it right. Could we merge this now and then I will work on resolving the conflicts?

WardBrian:
I was hoping @jsoules would take a look - he did already see a lot of it while it was in progress, so maybe it’s not necessary before merging

WardBrian:
Rebased on main and made some cosmetic renaming of the types and tests

gui/src/app/Stanc/stancWorker.ts
```diff
@@ -0,0 +1,48 @@
...
+  const { purpose, name, code } = e.data;
+
+  if (!stanc) {
+    self.postMessage({ error: "stanc.js not loaded!" });
+    return;
+  }
+
+  const args = [`filename-in-msg=${name}`, "auto-format", "max-line-length=78"];
+
+  const output = stanc(name, code, args);
```
jsoules, line 32: Would be nice to describe a little bit more the interface of the `stanc` entity (esp. since it's loaded as a code blob). I think I get it from looking at the usage, but I am insecure when dealing with ambiguity

gui/src/app/Stanc/useStanc.ts
```diff
@@ -0,0 +1,67 @@
+import { useCallback, useEffect, useState } from "react";
+import { StancReplyMessage, StancErrors, Replies, Requests } from "./Types";
+import stancWorkerURL from "./stancWorker?worker&url";
```
jsoules, line 3: remind me what these magic characters do? `?` means 'instantiate as a web worker' and `&url` means 'get the URL of the deployed asset' or something?

gui/src/app/Stanc/useStanc.ts
```diff
@@ -0,0 +1,67 @@
...
+        console.error(error);
+        return;
+      }
+      setStancErrors({ ...e.data });
+      if (purpose === Replies.Formatted) {
+        if (result) {
+          onFormat(result);
+        }
+      }
+    };
```
jsoules, line 43: Also a place where just a smidge more documentation of the API will go a long way to me understanding this in two weeks when I've forgotten what it does :sweat_smile:

gui/src/app/Stanc/Types.ts
```diff
@@ -0,0 +1,33 @@
...
+export enum Requests {
+  Format = "format",
+  Check = "check",
+}
+
+export type StancRequestMessage = {
+  purpose: Requests;
+  name: string;
+  code: string;
+};
```
jsoules, line 33: At the risk of being annoying--would you be willing to be a bit more specific with the names of the `Requests` and `Replies` enums? Like `StancWorkerRequests`?

Additionally, the `Format` and `Check` members could be more descriptive, when they appear elsewhere, as `SyntaxCheck` or `ReformatStanCode` or something like this. (I'm not attached to the specifics, just noticing that `Replies.Formatted` in `useStanc.ts` reads casually as looking at replies that are formatted, rather than the Reply that says "hey I formatted your code")

jsoules (commented):
Looks good. I've suggested a couple places for adding some comments or renaming to make things a little more obvious, but this is great on the substance.

Also, despite your misgivings in the PR writeup, I do think this was a useful thing to have done, both in the learning experience itself (for all of us) and in making the code more modular--that is a real improvement.

gui/src/app/FileEditor/StanCompileResultWindow.tsx
```diff
@@ -1,60 +1,33 @@
...
 
-    if (!model) return <div />
-    if ((model.errors) && (model.errors.length > 0)) {
+    if ((stancErrors.errors) && (stancErrors.errors.length > 0)) {
         return (
-            <div style={{width, height, color: 'red', padding: 0, overflow: 'auto'}}>
+            <div style={{ width, height, color: 'red', padding: 0, overflow: 'auto' }}>
                 <h3>Errors</h3>
-                {model.errors.map((error, i) => <div key={i} style={{font: 'courier', fontSize: 13}}><pre>{error}</pre></div>)}
+                {stancErrors.errors.slice(1).map((error, i) => <div key={i} style={{ font: 'courier', fontSize: 13 }}><pre>{error}</pre></div>)}
```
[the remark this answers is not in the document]
WardBrian: If you load up the `main` branch and introduce an error, you'll notice there is a random `0` on the page. This removes it.

I don't know why stancjs puts it there, and as far as I can tell several other places in the Stan ecosystem have adapted to just remove it. I don't like blaming other developers, but let's just say this behavior predates me working on the compiler.

Warnings doesn't have it. Go figure.

gui/src/app/Stanc/useStanc.ts
WardBrian: A `?worker` says "bundle the file for use in a web worker"
A `?url` says (as you guessed) "rather than import something, import _a url pointing to the deployed asset_"
A `?raw` means "import the inlined raw text of the file" (this is used in the worker itself)

All are documented at https://vitejs.dev/guide/assets

gui/src/app/Stanc/Types.ts
WardBrian: This and your above comment made me really think hard about what actually is needed, and I realized that there was some redundancy, so I cleaned a bunch up. Take another look?

gui/src/app/Stanc/stancWorker.ts
```diff
@@ -19,30 +24,37 @@ try {
   console.error(e);
 }
 
+// helper alias for type safety
+const postReply = (message: StancReplyMessage) => self.postMessage(message);
```
jsoules, line 28: also has the nice property of clarifying that you're responding to something. Elegant!

gui/src/app/Stanc/stancWorker.ts
```diff
@@ -19,30 +24,37 @@ try {
...
+  switch (purpose) {
+    case StancWorkerRequests.FormatStanCode: {
+      postReply(output);
+      break;
+    }
+    case StancWorkerRequests.CheckSyntax: {
+      // if we just syntax checked, don't send back formatted code
+      const { errors, warnings } = output;
+      postReply({ errors, warnings });
+      break
```
jsoules, line 57: Nice

gui/src/app/Stanc/useStanc.ts
jsoules: If you wouldn't object to just including the reference to that URL above, it'd be a service--the keywords are otherwise non-obvious and googling punctuation is notoriously fraught

jsoules (approved):
Looks great.

PR #70, opened 2024-06-14
Add minimal github action for yarn tests

WardBrian:
Now that we have tests, we can have our first little piece of CI


magland:
I think it's sufficiently self-contained to just merge.

PR #73, opened 2024-06-14
hide syntax error window by default

magland:
This replaces #51. Same functionality. Was just easier to start from scratch rather than resolve the conflicts caused by the other PRs.

### DIFF
--- a/gui/src/app/FileEditor/StanCompileResultWindow.tsx
+++ b/gui/src/app/FileEditor/StanCompileResultWindow.tsx
@@ -1,33 +1,45 @@
-import { Done } from "@mui/icons-material";
+import { Close, Done } from "@mui/icons-material";
 import { FunctionComponent } from "react";
 import { StancErrors } from "../Stanc/Types";
+import { SmallIconButton } from "@fi-sci/misc";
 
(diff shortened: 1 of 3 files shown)


WardBrian:
What does this do with warnings?

a good example is to name a variable “jacobian”. This is going to be a reserved word in a future version of Stan, so for now it gives a compiler warning

magland:
> What does this do with warnings?
> 
> a good example is to name a variable “jacobian”. This is going to be a reserved word in a future version of Stan, so for now it gives a compiler warning

Good point. Added a button for "Syntax warning"

image

magland:
@WardBrian @jsoules is this okay to merge?

jsoules (commented):
Perhaps dark orange might be a more appropriate (but still contrasting) color for warnings than blue?

Otherwise this seems fine to me if it's okay with you, @WardBrian

WardBrian:
Yep, this is fine by me!

I think we may eventually be able to replace this with something built in to the Monaco editor for showing errors (see #76 and https://github.com/WardBrian/vscode-stan-extension), but this is definitely an improvement over the current behavior

PR #74, opened 2024-06-17
[NO MERGE] prototypes for data.py and data.r

magland:
Replaces #26 and #27

Built on #65 because that provides a clean starting point for routes.

This provides prototypes for data.py and data.r data generation via pyodide and webr. These can be viewed in the page=prototypes route by clicking on the "view prototypes" link in the menu bar as shown:

image

You can select which prototype to view in the left panel

image

The data.py and data.r examples include example scripts for generating the data.json

### DIFF
--- a/gui/package.json
+++ b/gui/package.json
@@ -23,6 +23,7 @@
     "jszip": "^3.10.1",
     "monaco-editor": "^0.48.0",
     "plotly.js": "^2.33.0",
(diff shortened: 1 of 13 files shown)


magland:
Added analysis.py prototype that generates some plots in the output window.

(there's probably a better choice for the "analysis.py" file name)

image

PR #75, opened 2024-06-17
Clean up dependencies, use plotly reduced distribution

WardBrian:
This makes the change discussed [here](https://github.com/flatironinstitute/stan-playground/pull/47#discussion_r1631681184) to reduce the size of the plotly.js bundle by using a distribution with fewer features. For now I chose `cartesian`, which includes bar, box, contour, heatmap, histogram, histogram2d, histogram2dcontour, image, pie, scatter, scatterternary and violin. 

This reduces the size of the bundle from ~3.7mb to ~1.3mb. This is still lazy-loaded, just now it will lazily load faster.

I also removed some currently unused dependencies.


jsoules (approved):
This all seems pretty straightforward to me...

PR #80, opened 2024-06-25
Zip import/export (based from zip-import-export branch)

jsoules:
This is a modification of @magland's zip-import-export branch (PR #53) to use the amended data model from PR #77.
UI features are unchanged from that PR, so I'll refer to that for description, and this write-up will focus on the code changes (particularly as distinct from that PR/branch).

Most of the changes are to back-end files located under the `gui/src/app/SPAnalysis` directory. Front-end changes are under `gui/src/app/pages/HomePage/`. A few other smaller changes are in other directories and I'll describe them with full paths.

--
Probably the biggest focus of this PR is to establish standardized filesystem representations for the parts of the overall data model (`SPAnalysisDataModel`). This mapping is mainly implemented in **FileMapping.ts**. I've also created a new file for serialization code in **SPAnalysisSerialization.ts**. The reducer also had some significant changes to handle some new ways of updating the data model.

On the front-end side, code is mostly the same as in @magland's branch. I made some small changes to the **ExportWindow.tsx** and **ImportWindow.tsx** components to move some of their logic into the back-end code section (incl. the reducer), but as this is conceived as an update to that branch I tried to leave most of the functionality untouched.

Additionally, I've added `replaceSpaces.ts` and `triggerDownload.ts` functions in the `util` directory; these are small and straightforward.

Here are some notes on the back-end changes:
- [FileMapping.ts](https://github.com/flatironinstitute/stan-playground/blob/zip-import-export-js/gui/src/app/SPAnalysis/FileMapping.ts): Probably the trickiest part, but is commented pretty extensively.
  - Defines an enum for filesystem filenames (e.g. `main.stan`, `data.json`, etc.)
  - Defines a mapping from filenames to data-model field names.
  - Defines a type that maps filenames to filename contents.
  - Function to map a data model to a (partial) list of persistent files.
  - Function to take a (partial) list of persistent files and map their contents to a data model.
- [SPAnalysisReducer](https://github.com/flatironinstitute/stan-playground/blob/zip-import-export-js/gui/src/app/SPAnalysis/SPAnalysisReducer.ts):
  - Add a new action type that takes a partial list of data model fields & maps their contents to an instance of the data model.
  - Added functions to update data model metadata, data model sampling opts, and other fields (the "files" fields, which are text blobs).
  - All above functions have an option to determine whether or not to preserve any existing data that isn't being touched.
- [SPAnalysisSerialization](https://github.com/flatironinstitute/stan-playground/blob/zip-import-export-js/gui/src/app/SPAnalysis/SPAnalysisSerialization.ts): New file.
  - Move the "serialize everything to local storage" functions (which had been in the data model file).
  - Moved the zip serialization/deserialization code here from where it was in PR #53 (on the UI components).
- [SPAnalysisDataModel](https://github.com/flatironinstitute/stan-playground/blob/zip-import-export-js/gui/src/app/SPAnalysis/SPAnalysisDataModel.ts):
  - Added type exports, including one for the data model with the "ephemera" key omitted.
  - Added utility functions for copying permanent state to the ephemera, for returning appropriately stringified fields from the data model object, for looking up defined-field keys, and tracking whether models have been changed.
- [SPAnalysisContextProvider](https://github.com/flatironinstitute/stan-playground/blob/zip-import-export-js/gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx): trivial changes.


gui/package.json
```diff
@@ -25,6 +25,8 @@
     "plotly.js-cartesian-dist": "^2.33.0",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-dropzone": "^14.2.3",
+    "react-markdown": "^8",
```
WardBrian, line 29: Does this PR actually depend on markdown?

gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
```diff
@@ -42,17 +44,26 @@ export const initialDataModel: SPAnalysisDataModel = {
...
 }
 
-export const deserializeAnalysis = (serialized: string): SPAnalysisDataModel => {
-    const intermediary = JSON.parse(serialized)
-    // Not sure if this is strictly necessary
-    intermediary.ephemera = {}
-    const stringFileKeys = Object.values(SPAnalysisKnownFiles).filter((v) => isNaN(Number(v)));
-    stringFileKeys.forEach((k) => intermediary.ephemera[k] = intermediary[k]);
-    return intermediary as SPAnalysisDataModel
+export const getStringKnownFileKeys = () => Object.values(SPAnalysisKnownFiles).filter((v) => isNaN(Number(v)));
```
WardBrian, line 56: What is this?

gui/src/app/util/triggerDownload.ts
```diff
@@ -0,0 +1,10 @@
+
+export const triggerDownload = (blob: Blob, name: string, onClose: () => void) => {
```
WardBrian, line 2: This can also be used in SamplerOutputViews, I believe

gui/src/app/util/replaceSpaces.ts
```diff
@@ -0,0 +1,3 @@
+export const replaceSpaces = (str: string) => {
```
WardBrian, line 1:
```suggestion
export const replaceSpacesWithUnderscores = (str: string) => {
```

gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
```diff
@@ -9,7 +9,7 @@ type SPAnalysisFiles = {
     [filetype in SPAnalysisKnownFiles]: string
 }
 
-type SPAnalysisBase = SPAnalysisFiles &
+export type SPAnalysisBase = SPAnalysisFiles &
```
WardBrian, line 12: Are we actually using the fact that this is exported somewhere?

WardBrian (commented):
A few misc comments/questions

gui/package.json
jsoules: This might've been a mixup in the rebase.

gui/src/app/util/triggerDownload.ts
jsoules: I knew it'd come in handy somewhere :)

gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
jsoules: I don't think so--probably an artifact.

magland:
Tried in a codespace and cot this error

```
 [ERROR] No matching export in "src/app/util/replaceSpaces.ts" for import "replaceSpaces"

    src/app/SPAnalysis/SPAnalysisSerialization.ts:2:9:
      2 │ import { replaceSpaces } from "../util/replaceSpaces"
        ╵          ~~~~~~~~~~~~~


    at failureErrorWithLog (/workspaces/stan-playground/gui/node_modules/esbuild/lib/main.js:1472:15)
    at /workspaces/stan-playground/gui/node_modules/esbuild/lib/main.js:945:25
    at runOnEndCallbacks (/workspaces/stan-playground/gui/node_modules/esbuild/lib/main.js:1315:45)
    at buildResponseToResult (/workspaces/stan-playground/gui/node_modules/esbuild/lib/main.js:943:7)
    at /workspaces/stan-playground/gui/node_modules/esbuild/lib/main.js:955:9
    at new Promise (<anonymous>)
    at requestCallbacks.on-end (/workspaces/stan-playground/gui/node_modules/esbuild/lib/main.js:954:54)
    at handleRequest (/workspaces/stan-playground/gui/node_modules/esbuild/lib/main.js:647:17)
    at handleIncomingPacket (/workspaces/stan-playground/gui/node_modules/esbuild/lib/main.js:672:7)
    at Socket.readFromStdout (/workspaces/stan-playground/gui/node_modules/esbuild/lib/main.js:600:7)
```

I think  there was a function name change?

jsoules:
> Tried in a codespace and got this error
> [...]
> I think there was a function name change?

Yep, I committed the change direct from the comment and missed that it needed to be corrected where it is imported as well. Should be fixed now.

gui/src/app/util/triggerDownload.ts
jsoules: Pushed an update to use it there as well.

gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
jsoules: The enum identifies the actual names of the fields on the object (which are enforced by the type system). This is a way to get the keys out of the enum so that we can access what those field names are. It's used in the serialization code (SPAnalysisSerialization) to iterate over the field names defined by the `SPAnalysisKnownFiles` type.

...we really should change that name since they aren't actually files per se...

WardBrian: I guess my question is why is NAN involved. Just a weird JS thing?

jsoules: OH, sorry. Yeah, that's a bit of an artifact--while I don't think it applies in this case, by default Typescript provides a reverse mapping for enums so that you can go from the keys back to the values. I believe this only works for numeric-valued enums (a lot of this code is specifically trying to solve the problem of a reverse mapping doesn't exist when you're using non-numeric enums) but when iterating over the enum members, you often want to filter out numeric values to avoid funny results.

That said, this is relevant for 1) numeric-valued enums and 2) when looping over their keys, so it can probably be removed!

jsoules: (https://blog.logrocket.com/iterate-over-enums-typescript/ for some background, but again this is probably just pleonastic code in this situation)

jsoules: or https://www.typescriptlang.org/play/?target=10&module=1#code/KYDwDg9gTgLgBMAdgVwLZwLIE8CiL0DeAUHHAIJwC8cA5AGYCWUAzjDQDQlwBCVtzwAMYREAEw5cAwnwCsRAL5EiAeQBGAKyEwAdADcAhgBtkwZgApseNAEptdaDn2CAFmbMBra1QB8cYYmYIQ2BtQwgAczMAAwBpOAZmOAASAnd5KOtrJSA for some of the funky behavior you can get with the auto-reverse mapping

WardBrian: Thanks for the links

> so it can probably be removed!

Can it?

gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
```diff
@@ -42,17 +44,26 @@ export const initialDataModel: SPAnalysisDataModel = {
...
-    // Not sure if this is strictly necessary
-    intermediary.ephemera = {}
-    const stringFileKeys = Object.values(SPAnalysisKnownFiles).filter((v) => isNaN(Number(v)));
-    stringFileKeys.forEach((k) => intermediary.ephemera[k] = intermediary[k]);
-    return intermediary as SPAnalysisDataModel
+export const getStringKnownFileKeys = () => Object.values(SPAnalysisKnownFiles).filter((v) => isNaN(Number(v)));
+
+export const modelHasUnsavedChanges = (data: SPAnalysisDataModel): boolean => {
+    const stringFileKeys = getStringKnownFileKeys()
+    return stringFileKeys.every((k) => data[k] !== data.ephemera[k])
```
WardBrian, line 60: Should this be `some` instead of `every`?

jsoules: It sure should...

gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
```diff
@@ -42,17 +44,26 @@ export const initialDataModel: SPAnalysisDataModel = {
     samplingOpts: defaultSamplingOpts
 }
 
-export const serializeAnalysis = (data: SPAnalysisDataModel): string => {
-    const intermediary = {
-        ...data, ephemera: undefined }
-    return JSON.stringify(intermediary)
+export const persistStateToEphemera = (data: SPAnalysisDataModel): SPAnalysisDataModel => {
```
WardBrian, line 47: [note to self] this is one of the helpers I wanted in #84 and I should use it after this is merged

gui/src/app/pages/HomePage/ImportWindow.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }, [filesUploaded, onClose, update])
+
+    useEffect(() => {
+        if (!filesUploaded) return
+        if ((filesUploaded.length === 1) && (!filesUploaded[0].name.endsWith('.zip'))) {
+            // The user has uploaded a single file and it is not a zip file. In
+            // this case we want to give the user the option whether or not to
+            // replace the current project.
+            setShowReplaceProjectOptions(true)
+        }
```
WardBrian, line 60: Happy to leave this until after merge, just echoing my discussion from #53 that I think uploading a single file should keep the drop box enabled to allow you to also add a second/third etc

gui/yarn.lock
WardBrian, line 1: Note: when I run `yarn` locally, I end up with changes in my yarn.lock compared to this branch

WardBrian (approved):
This looks good to me! I left a couple comments, none of which are blockers on a merge IMO

PR #82, opened 2024-06-25
Reintroduce formatting for stanie data file

WardBrian:
While reviewing #80 I noticed that the Stan example data was being loaded in all on one long line


magland:
This looks good, but a side note about "stanie"...

... this was meant to be a temporary way to hard-code some examples. I think we should host the examples in gists instead and remove the "stanie" stuff once that's in place.

Issue #81, opened 2024-06-25
Use stanard React Router over custom useRoute


jsoules:
I thought I had an issue for this already, but it looks like not, I guess we only discussed in person.

In reviewing PR #65 as part of the discussion around PR #77, I realized how much the `useRoute` infrastructure will need to grow to support query parsing.

I think we will be much better off moving away from custom routing logic implementation and using stock React Router functionality (both the `use-search-params` hook, as well as more standard route structuring); & I think this should be a part of the query parameter support PR. I have a model for this in the QUASR Navigator at https://github.com/jsoules/stellarator-navigator/blob/trunk/stellarator-navigator/src/main.tsx and happy to help with it.

magland:
Sounds good, but before fussing with the implementation, I think it would be helpful to make a design spec for exactly what the url query parameters should look like.

Would it deviate from this?
https://github.com/flatironinstitute/stan-playground/pull/65#issue-2351527832

WardBrian:
I don’t think the URL structure will differ from what we had previously discussed there

jsoules:
> Sounds good, but before fussing with the implementation, I think it would be helpful to make a design spec for exactly what the url query parameters should look like.
> 
> Would it deviate from this? [#65 (comment)](https://github.com/flatironinstitute/stan-playground/pull/65#issue-2351527832)

I agree that that URL structure is fine, though I'd also be happy to allow both sampling_opts and inline_opts with a last-one-wins (or query string beats other source). No need to worry over that right now--should be easy enough to change it later if we want.

WardBrian:
Issue https://github.com/flatironinstitute/stan-playground/issues/50 seems to cover the re-implementation better

PR #83, opened 2024-06-25
Remove existing router logic

WardBrian:
Related: #81.

Before taking a look at implementations of the query parameters on the new data model, I wanted to start from a cleaner slate. This should be feature equivalent, since we were not really doing any routing yet


magland:
I'm in favor of this as long as we get a replacement in there relatively soon so that it doesn't block loading from query parameters and loading from gists.

@jsoules ?

jsoules:
I have no objection to this.

Issue #50, opened 2024-06-10
proposed functionality for loading/saving analyses


magland:
@WardBrian @jsoules

Below are various possible conventions for loading/saving SP analyses/projects

**Specify individual files in URL query parameters**

`https://stan-playground.vercel.app?...query...`

```
main.stan - URL to remote main.stan file
data.json - URL to remote data.json file
sampling_opts.json - URL to remote sampling_opts.json file
t - title

FUTURE:
data.py - URL to remote data.py file for generating data.json
data.r - URL to remote data.r file for generating data.json
analyze.py (or similar) - URL to remote analyze.py file for downstream analysis of sampling output - generating figures, tables, etc
analyze.r (or similar) - same, except for r language

POSSIBLE FUTURE:
samples.csv
```

**Specify individual files in URL query parameters (different naming) from #39**

`https://stan-playground.vercel.app?...query...`

```
stanURL - URL to remote main.stan file
dataURL - URL to remote data.json file
samplingOptsURL - URL to remote sampling_opts.json file
t - title
project (or projectURL) - URL to a project.json file that contains any/all of the above

FUTURE:
dataPyURL
dataRURL
analyzePyURL
analyzeRURL
```

**Load from a GitHub Gist**

```
`https://stan-playground.vercel.app?project=https://gist.github.com/magland/da3d4143276827609ec9317bb3db8b04`
```

The Gist is a directory of files including `main.stan`, `data.json`, `sampling_opts.json`, etc.

The description of the Gist is the title

**Load from a single project file, reference file system**

```
`https://stan-playground.vercel.app?project=[projectURL]`
```

Points to a remote json file that is a reference file system that includes `main.stan`, etc.

More information about reference file system

https://fsspec.github.io/kerchunk/spec.html

Here's an example rfs

```json
{
    "version": 1,
    "meta": {
        "title": "..."
    },
    "refs": {
      "main.stan": "<the STAN project text>",
      "data.json": "...",
      "sampling_opts.json": "..."
    }
}
```

**Importantly you can also do this** (this is why it's called a "reference file system")

```json
{
    "version": 1,
    "meta": {
        "title": "..."
    },
    "refs": {
      "main.stan": [stanURL],
      "data.json": [dataURL],
      "sampling_opts.json": "..."
    }
}
```

So you can mix and max inline/external references

**Save to new Gist**

Save the current analysis/project as a new Gist with the above convention

**Update an existing Gist**

[...]

PR #39, opened 2024-06-06
Minimal support for loading from URL query parts

WardBrian:
This is an alternative to #35 for discussion.

There are five possible query parameters at the moment:

- `stanURL` - where to fetch the stan code from
- `dataURL` - ditto but for data JSON
- `samplingOptsURL`  - ditto but for sampler options
- `t` - title
- `project` - a bit special, this one points to a JSON which _contains any/all of the above_

Here's an example (assuming you are running locally):

http://127.0.0.1:3000/?stanURL=https://raw.githubusercontent.com/stan-dev/example-models/master/misc/eight_schools/eight_schools.stan&dataURL=https://raw.githubusercontent.com/stan-dev/example-models/master/misc/eight_schools/eight_schools.data.json&t=Eight%20Schools

Here's the same content but wrapped up in one JSON:
http://127.0.0.1:3000/?project=https://gist.githubusercontent.com/WardBrian/b6ab1f73d9c45ae43b785b1752a9966b/raw/16e17f1c8b85d4a29ceba3475d695c771e28d870/eightschools.json

This also seems like it is a prime candidate for a further refactor into something like a reducer, but for now I thought it was clearer to see what changed if I held off on anything further.

PR #84, opened 2024-06-26
Basic query loading

WardBrian:
This is my combination of #65 and #39 on top of the new data model and react-router-dom's useSearchParams.

Jeremy's example link of 

http://127.0.0.1:3000/?stan=https://gist.githubusercontent.com/magland/da3d4143276827609ec9317bb3db8b04/raw/cf48846a36de0559f96a9e4bcb5a5dcbe46383be/main.stan&data=https://gist.githubusercontent.com/magland/da3d4143276827609ec9317bb3db8b04/raw/cf48846a36de0559f96a9e4bcb5a5dcbe46383be/data.json&num_chains=6&num_warmup=1000&num_samples=1000&init_radius=2&seed=6&title=test_title

still works in this PR. 

This is the first part of #50 

Wishlist:
1. This added a lot of code to the ContextProvider which may be better put elsewhere
2. I noted a few places where utility functions may be useful, but some I think already exist in other PRs (I already took @jsoules' `modelHasUnsavedChanges`

### DIFF
--- a/gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
+++ b/gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
@@ -1,5 +1,5 @@
-import { createContext, FunctionComponent, PropsWithChildren, useEffect, useReducer } from "react"
-import {  initialDataModel, SPAnalysisDataModel, modelHasUnsavedChanges } from "./SPAnalysisDataModel"
+import { initialDataModel, SPAnalysisDataModel } from "./SPAnalysisDataModel"
+import { createContext, FunctionComponent, PropsWithChildren, useCallback, useEffect, useReducer } from "react"
 import { SPAnalysisReducer, SPAnalysisReducerAction, SPAnalysisReducerType } from "./SPAnalysisReducer"
 import { useSearchParams } from "react-router-dom"
 import { deserializeAnalysisFromLocalStorage, serializeAnalysisToLocalStorage } from "./SPAnalysisSerialization"
@@ -20,9 +20,18 @@ export const SPAnalysisContext = createContext<SPAnalysisContextType>({
 
 
 const SPAnalysisContextProvider: FunctionComponent<PropsWithChildren<SPAnalysisContextProviderProps>> = ({ children }) => {
-    const [data, update] = useReducer<SPAnalysisReducerType>(SPAnalysisReducer, initialDataModel)
-
     const [searchParams, setSearchParams] = useSearchParams();
+
+    const onDirty = useCallback(() => {
+        // whenever the data state is 'dirty', we want to
+        // clear the URL bar as to indiciate that the viewed content is
+        // no longer what the link would point to
+        if (searchParams.size !== 0)
+            setSearchParams(new URLSearchParams())
+    }, [searchParams, setSearchParams]);
+
+
+    const [data, update] = useReducer<SPAnalysisReducerType>(SPAnalysisReducer(onDirty), initialDataModel)
 
     ////////////////////////////////////////////////////////////////////////////////////////
     // For convenience, we save the state to local storage so it is available on
@@ -62,17 +71,6 @@ const SPAnalysisContextProvider: FunctionComponent<PropsWithChildren<SPAnalysisC
         }
 
     }, [data, searchParams])
-
-
-    useEffect(() => {
-        // whenever the data state is 'dirty', we want to
-        // clear the URL bar as to indiciate that the viewed content is
-        // no longer what the link would point to
-        if (searchParams.size !== 0 && modelHasUnsavedChanges(data)) {
-            setSearchParams(new URLSearchParams());
-        }
-    }, [data, searchParams, setSearchParams])
-
 
     return (
         <SPAnalysisContext.Provider value={{ data, update }}>
--- a/gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
+++ b/gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
@@ -1,9 +1,8 @@
 import { initialDataModel, SPAnalysisDataModel } from "./SPAnalysisDataModel"
-import { createContext, FunctionComponent, PropsWithChildren, useCallback, useEffect, useReducer } from "react"
+import { createContext, FunctionComponent, PropsWithChildren, useEffect, useReducer } from "react"
 import { SPAnalysisReducer, SPAnalysisReducerAction, SPAnalysisReducerType } from "./SPAnalysisReducer"
-import { useSearchParams } from "react-router-dom"
 import { deserializeAnalysisFromLocalStorage, serializeAnalysisToLocalStorage } from "./SPAnalysisSerialization"
-import { fromSearchParams, fetchRemoteAnalysis, queryStringHasParameters } from "./SPAnalysisQueryLoading"
+import { fetchRemoteAnalysis, queryStringHasParameters, useQueryParams } from "./SPAnalysisQueryLoading"
 
 type SPAnalysisContextType = {
     data: SPAnalysisDataModel
@@ -20,18 +19,10 @@ export const SPAnalysisContext = createContext<SPAnalysisContextType>({
 
 
 const SPAnalysisContextProvider: FunctionComponent<PropsWithChildren<SPAnalysisContextProviderProps>> = ({ children }) => {
-    const [searchParams, setSearchParams] = useSearchParams();
 
-    const onDirty = useCallback(() => {
-        // whenever the data state is 'dirty', we want to
-        // clear the URL bar as to indiciate that the viewed content is
-        // no longer what the link would point to
-        if (searchParams.size !== 0)
-            setSearchParams(new URLSearchParams())
-    }, [searchParams, setSearchParams]);
+    const { queries, clearSearchParams } = useQueryParams();
 
-
-    const [data, update] = useReducer<SPAnalysisReducerType>(SPAnalysisReducer(onDirty), initialDataModel)
+    const [data, update] = useReducer<SPAnalysisReducerType>(SPAnalysisReducer(clearSearchParams), initialDataModel)
 
     useEffect(() => {
         // as user reloads the page or closes the tab, save state to local storage
@@ -49,9 +40,8 @@ const SPAnalysisContextProvider: FunctionComponent<PropsWithChildren<SPAnalysisC
     useEffect(() => {
         if (data != initialDataModel) return;
 
-        const query = fromSearchParams(searchParams);
-        if (queryStringHasParameters(query)) {
-            fetchRemoteAnalysis(query).then((data) => {
+        if (queryStringHasParameters(queries)) {
+            fetchRemoteAnalysis(queries).then((data) => {
                 update({ type: 'loadInitialData', state: data })
             })
         } else {
@@ -62,7 +52,7 @@ const SPAnalysisContextProvider: FunctionComponent<PropsWithChildren<SPAnalysisC
             update({ type: 'loadInitialData', state: parsedData })
         }
 
-    }, [data, searchParams])
+    }, [data, queries])
 
     return (
         <SPAnalysisContext.Provider value={{ data, update }}>
--- a/gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
+++ b/gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
@@ -1,6 +1,8 @@
 import { createContext, FunctionComponent, PropsWithChildren, useEffect, useReducer } from "react"
-import { initialDataModel, SPAnalysisDataModel } from "./SPAnalysisDataModel"
+import {  initialDataModel, SPAnalysisDataModel, modelHasUnsavedChanges } from "./SPAnalysisDataModel"
 import { SPAnalysisReducer, SPAnalysisReducerAction, SPAnalysisReducerType } from "./SPAnalysisReducer"
+import { useSearchParams } from "react-router-dom"
+import { SamplingOpts } from "../StanSampler/StanSampler"
 import { deserializeAnalysisFromLocalStorage, serializeAnalysisToLocalStorage } from "./SPAnalysisSerialization"
 
 type SPAnalysisContextType = {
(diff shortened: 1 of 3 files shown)


gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
```diff
@@ -12,12 +14,113 @@ type SPAnalysisContextProviderProps = {
...
+const deepCopy = (obj: any) => {
+    return JSON.parse(JSON.stringify(obj))
+}
+
+const fetchRemoteAnalysis = async (query: QueryParams) => {
+
+    // any special 'project' query could be loaded here at the top
+    const data: SPAnalysisDataModel = deepCopy(initialDataModel);
+
+    if (query.stan) {
```
magland, line 78: We could save this for later, but I think it would be better to iterate over the query parameters in this kind of fashion

```
for key in query
  if key == QueryParamKeys.StanFile
    data.stan = ...
  elif key == ...
    ..
  else
    print a warning that the query parameter was unused
```

The advantages: we'd get the warnings; we'd make use of the QueryParamKeys enum
```

File: gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx [hunk 2]
```diff
@@ -35,17 +138,37 @@ const SPAnalysisContextProvider: FunctionComponent<PropsWithChildren<SPAnalysisC
...
+
     useEffect(() => {
-        // load the saved state on first load
-        const savedState = localStorage.getItem('stan-playground-saved-state')
-        if (!savedState) return
-        const parsedData = deserializeAnalysis(savedState)
-        update({ type: 'loadLocalStorage', state: parsedData })
-    }, [])
+        if (searchParams.size !== 0 && modelHasUnsavedChanges(data)) {
+            setSearchParams(new URLSearchParams());
```
magland, line 173: Would be good to leave a comment here explaining that we are clearing the query string, and why we are doing that.

magland (approved):
Aside from my two minor suggestions this looks great.

gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
WardBrian: Good idea. I added a warning inside the `fromSearchParams` function (just to the console for now), since by the time we have this `query` object it would only ever have keys we knew about

WardBrian:
Ok, I noticed the URL wasn't clearing when you would click on a different example on the sidebar. I have a working solution, but it required the reducer to accept in a `onDirty` function, which I'm not entirely sure is idiomatic?

@jsoules

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    if (query.title) {
+        data.meta.title = query.title
+    }
+
+    // TODO(bmw) create a 'setEphemera' helper?
+    data.ephemera.stanFileContent = data.stanFileContent
+    data.ephemera.dataFileContent = data.dataFileContent
```
jsoules, line 110: as discussed, this is in the zip PR

gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
```diff
@@ -12,11 +14,23 @@ type SPAnalysisContextProviderProps = {
...
+    const onDirty = useCallback(() => {
+        // whenever the data state is 'dirty', we want to
+        // clear the URL bar as to indiciate that the viewed content is
+        // no longer what the link would point to
+        if (searchParams.size !== 0)
+            setSearchParams(new URLSearchParams())
+    }, [searchParams, setSearchParams]);
+
+
+    const [data, update] = useReducer<SPAnalysisReducerType>(SPAnalysisReducer(onDirty), initialDataModel)
```
jsoules, line 33: I don't mind the `onDirty` pattern here; I haven't seen it before but it's a clever solution and pretty straightforward.

I would put the definition of the function in the SPAnalysisQueryLoading file though, since it's more to do with URL-munging, if it's possible to do so. (This might involve moving the `useSearchParams` hook out, or possibly just having multiple components use such a hook; it clearly has to be a singleton, so it's probably more in line with ` context` than a `state` whatever the documentation says. Or we could turn `fromSearchParams` into a hook defined in that file...)

gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
```diff
@@ -34,18 +48,31 @@ const SPAnalysisContextProvider: FunctionComponent<PropsWithChildren<SPAnalysisC
             window.removeEventListener('beforeunload', handleBeforeUnload);
         };
     }, [data])
+    ////////////////////////////////////////////////////////////////////////////////////////
```
jsoules, line 51: Wondering if we still need to block out this section of the code--seems like we might be happy sticking with this behavior?

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
```diff
@@ -0,0 +1,113 @@
...
+        sampling_opts: searchParams.get(QueryParamKeys.SamplingOpts),
+        title: searchParams.get(QueryParamKeys.Title),
+        num_chains: searchParams.get(QueryParamKeys.SONumChains),
+        num_warmup: searchParams.get(QueryParamKeys.SONumWarmup),
+        num_samples: searchParams.get(QueryParamKeys.SONumSamples),
+        init_radius: searchParams.get(QueryParamKeys.SOInitRadius),
+        seed: searchParams.get(QueryParamKeys.SOSeed),
+    }
+    return query
+
```
jsoules, line 40: stray newline?

gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
```diff
@@ -34,18 +48,31 @@ const SPAnalysisContextProvider: FunctionComponent<PropsWithChildren<SPAnalysisC
...
-        const parsedData = deserializeAnalysis(savedState)
-        update({ type: 'loadLocalStorage', state: parsedData })
-    }, [])
-    ////////////////////////////////////////////////////////////////////////////////////////
+        if (data != initialDataModel) return;
+
+        const query = fromSearchParams(searchParams);
+
+        // any query is set
+        if (Object.values(query).some(v => v !== null)) {
```
jsoules, line 60: Might be worth wrapping this in a semantically-named function `queryStringHasParameters: QueryParams -> boolean`--not super important, it'd still be a one-liner, but it would read a little more obviously than having to comment it (& would let you keep the definition of what a 'meaningful query string' is closer to the code that handles the query strings)

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
```diff
@@ -0,0 +1,113 @@
...
+        return undefined
+    }
+}
+
+const deepCopy = (obj: any) => {
+    return JSON.parse(JSON.stringify(obj))
+}
+
+export const fetchRemoteAnalysis = async (query: QueryParams) => {
+
```
jsoules, line 63: stray newline?

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
```diff
@@ -0,0 +1,113 @@
...
+    // any special 'project' query could be loaded here at the top
+    const data: SPAnalysisDataModel = deepCopy(initialDataModel)
+
+    if (query.stan) {
```
jsoules, line 67: Noting several `await`s here--there's always an asterisk in my head to want to make sure we aren't demanding that these be serialized when the fetches could be happening in parallel.

From a *very brief* perusal I wonder if we actually want to defer the `await`s until after all the `fetch`s have been triggered?
...sorry, that's a little gnomic, let me know if you'd like me to expand on that thought.

jsoules (commented):
Suggested moving some code around, a couple spurious formatting things, etc.

Probably the only substantial suggestion is that I think we might want to restructure the async stuff in the `FetchRemoteAnalysis` function--by eye it looks like a multi-part query string will trigger a bunch of serial fetches, awaiting between each fetch, rather than doing all the fetches in parallel. That won't be super noticeable locally, but might be perceptible for someone with a slower internet connection.

gui/src/app/SPAnalysis/SPAnalysisReducer.ts
```diff
@@ -23,13 +23,20 @@ export type SPAnalysisReducerAction = {
...
     state: SPAnalysisDataModel
 } | {
     type: 'clear'
 }
 
-export const SPAnalysisReducer: SPAnalysisReducerType = (s: SPAnalysisDataModel, a: SPAnalysisReducerAction) => {
+export const SPAnalysisReducer = (onDirty: () => void) => (s: SPAnalysisDataModel, a: SPAnalysisReducerAction) => {
+    if (a.type !== "loadInitialData") {
+        // TextEditor seems to trigger occasional spurious edits where nothing changes
+        if (a.type !== "editFile" || s[a.filename] != a.content) {
```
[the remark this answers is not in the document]
WardBrian: > Or maybe the desired behavior is just "clear the URL bar if the user breathes on anything" and we don't need to worry about this.

This is the behavior I would argue for. Even editing the sampler parameters when those weren't set in the url does break the rule of "is this the page I would view if I clicked on the link again"

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
WardBrian: Is 22aaba7fe72c41ba06a3627535ead1c11a64d142 what you meant? This isn't something I had thought about before but it makes sense

jsoules: Yeah, this seems more like.

gui/src/app/SPAnalysis/SPAnalysisReducer.ts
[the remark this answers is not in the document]
jsoules: Well, at that point it's kind of just indeterminate--if there's any part of the project that isn't set, you can't guarantee that you'll get the unset parts correct when you load, because they just aren't there. (This is more of a philosophical point right now, I'm not advocating for any particular practical decision as a result of this observation)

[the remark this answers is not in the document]
WardBrian: It would be the default value still, which is well-defined even if it isn't necessarily something we're guaranteeing will never change.

I agree something here could be more sophisticated, but for now it not being so is intentional

gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
WardBrian: I turned fromSearchParams into a hook in 01ced77695b43fce0600e5f226380363a8b297ca, good suggestion

gui/src/app/SPAnalysis/SPAnalysisReducer.ts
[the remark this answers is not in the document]
jsoules: Ahhh, okay. Confirmed that it does use the default values, not the last values.

magland:
Tried this out.

1. I think if the user edits a file but does not save local edits, the url query should not disappear, but should stay until the save button is actually pressed. Upon reloading page, if there are unsaved edits, the user should be warned with a popup. I don't see any downsides to this, and the upside is that user won't kill the query if they accidentally type something.

2. The back button is not working as I would expect. If I make changes and the url query disappears... then I use the back button it reappears, but the content does not revert. In my opinion it should revert - this would be a value add feature. To achieve this, it will be important to restructure a bit the relationship between the state and the route. Ideally the route should be the source of truth, and the state would flow from it. Obviously when they query is missing altogether that's a separate case and the source of truth needs to shift to a different entity.

3. Ideally, if a the sampling opts were provided in the url and the user modifies one of these opts, I think it would be best to keep the url intact and to just adjust the relevant part. This goes along with the route being the source of truth for the state.

4. It will be important to not only keep track of the content of the files, but also the source of the files (when unedited)... because when a user selects "generate url" in the future, and the files were unedited from their online versions, we need to be able to repopulate that.

That's a lot to consider... maybe these adjustments should wait for a future iteration.

WardBrian:
Item 1 is relatively simple, just changing the circumstances under which onDirty is called in the reducer. The pop up is a bit more involved, so I would prefer to leave this. 

For item 2, I would need to look into the back button more but I agree on what the behavior should be in the ideal world. 

I disagree on items 3 & 4. In particular, if someone sends me a link, I edit one file but not others, and I click share, I would very much like an entirely new copy to be made. That way, someone editing the sources of their original link does not make changes to mine unbeknownst to me

magland:
> Item 1 is relatively simple, just changing the circumstances under which onDirty is called in the reducer. The pop up is a bit more involved, so I would prefer to leave this.
> 
> For item 2, I would need to look into the back button more but I agree on what the behavior should be in the ideal world.
> 
> I disagree on items 3 & 4. In particular, if someone sends me a link, I edit one file but not others, and I click share, I would very much like an entirely new copy to be made. That way, someone editing the sources of their original link does not make changes to mine unbeknownst to me

Okay, let's not let 3 and 4 block this PR.

WardBrian:
Item 2 (back button functioning as expected) seems to be a result of preserving the URL until something is dirty. Anything I did that detected when the URL parameters changed (e.g. from the back button) led to either rendering the Example buttons on the side unusable due to the current url always overwriting them, or led to an infinite loop in the renderer. 

If it is implemented like I originally did way back in #39, where the URL is cleared as soon as loading is complete (the way godbolt.org works, for example), then it is "easy".

See diff here: https://github.com/flatironinstitute/stan-playground/compare/query-url-loading...query-url-loading-back-button

This also avoids the onDirty trickery in the reducer, so I prefer it, but it is a different feature set than the one we had agreed on a couple weeks ago. 

I would also be fine revisiting this in a later PR, if we'd prefer that. @magland @jsoules

magland:
> I would also be fine revisiting this in a later PR, if we'd prefer that. @magland @jsoules

I agree, this is quite tricky. I'd vote to leave it as is for now (where back button restores query but does not reload things)

jsoules:
I'm inclined toward just clearing the URL bar if it's causing problems, but we can worry about that later.

I'll merge this as is for now.

PR #90, opened 2024-06-26
include sampling_opts.json in exported csv zip with draws

### LINKED ISSUE #72: When downloading draws in zip, also save sampling_opts.json
See

_Originally posted by @WardBrian in https://github.com/flatironinstitute/stan-playground/pull/68#discussion_r1640365239_

magland:
Fixes #72 

Had to add `#samplingOpts` data member to StanSampler class.

### DIFF
--- a/gui/src/app/StanSampler/StanSampler.ts
+++ b/gui/src/app/StanSampler/StanSampler.ts
@@ -29,9 +29,8 @@ class StanSampler {
     #draws: number[][] = [];
     #computeTimeSec: number | undefined = undefined;
     #paramNames: string[] = [];
-    #numChains: number = 0;
     #samplingStartTimeSec: number = 0;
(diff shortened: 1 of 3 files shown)


gui/src/app/StanSampler/StanSampler.ts
```diff
@@ -31,6 +31,7 @@ class StanSampler {
     #paramNames: string[] = [];
     #numChains: number = 0;
     #samplingStartTimeSec: number = 0;
+    #samplingOpts: SamplingOpts | undefined = undefined; // the sampling options used in the last sample call
```
WardBrian, line 34: Now that we're storing numChains is redundant, correct?

I think we should set this to defaultSamplingOpts rather than undefined, and remove numChains

magland: Done.

PR #93, opened 2024-06-26
type guards for SPAnalysisDataModel

### LINKED ISSUE #89: type guard needed when loading from localStorage
It's important to check that the state loaded from local browser storage conforms to the SPAnalysisDataModel type.

I ran into a problem where I had old content in my

```
const savedState = localStorage.getItem('stan-playground-saved-state')
```

And it caused the page to crash. I had to manually clear the local storage.

In `deserializeAnalysisFromLocalStorage` we should have a type guard instead of forced casting:

```
return intermediary as SPAnalysisDataModel
```

So we should do something like

```
if (!isSPAnalysisDataModel(intermediary)) {
  console.warn('...')
  return null
}
return intermediary
```

Then handle null appropriately.

magland:
Fixes #89 

Added some type guards which are used to check the input of local storage and also in `loadFromProjectFiles`

This was somewhat difficult to test because if you manually put an error in the local storage, that gets overwritten upon page reload. So I managed to achieve this test by putting a temporary error in the write-to-local-storage code and verified that the page did not crash and I got the warning message upon reload.

### DIFF
--- a/gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
+++ b/gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
@@ -27,6 +27,7 @@ const isSPAnalysisBase = (x: any): x is SPAnalysisBase => {
     if (!x) return false
     if (typeof x !== 'object') return false
     if (!isSamplingOpts(x.samplingOpts)) return false
+    if (!isSPAnalysisFiles(x)) return false
     return true
 }
 
--- a/gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
+++ b/gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
@@ -1,4 +1,4 @@
-import { SamplingOpts, defaultSamplingOpts } from "../StanSampler/StanSampler"
+import { SamplingOpts, defaultSamplingOpts, isSamplingOpts } from "../StanSampler/StanSampler"
 
 export enum SPAnalysisKnownFiles {
     STANFILE = 'stanFileContent',
@@ -9,13 +9,36 @@ type SPAnalysisFiles = {
     [filetype in SPAnalysisKnownFiles]: string
 }
 
+const isSPAnalysisFiles = (x: any): x is SPAnalysisFiles => {
+    if (!x) return false
+    if (typeof x !== 'object') return false
+    for (const k of Object.values(SPAnalysisKnownFiles)) {
+        if (typeof x[k] !== 'string') return false
+    }
+    return true
+}
+
 type SPAnalysisBase = SPAnalysisFiles &
 {
     samplingOpts: SamplingOpts
 }
 
+const isSPAnalysisBase = (x: any): x is SPAnalysisBase => {
+    if (!x) return false
+    if (typeof x !== 'object') return false
+    if (!isSamplingOpts(x.samplingOpts)) return false
+    return true
+}
+
 type SPAnalysisMetadata = {
     title: string
+}
+
+export const isSPAnalysisMetaData = (x: any): x is SPAnalysisMetadata => {
+    if (!x) return false
+    if (typeof x !== 'object') return false
+    if (typeof x.title !== 'string') return false
+    return true
 }
 
 type SPAnalysisEphemeralData = SPAnalysisFiles & {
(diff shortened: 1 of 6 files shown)


gui/src/app/SPAnalysis/SPAnalysisDataModel.ts
```diff
@@ -9,28 +9,66 @@ type SPAnalysisFiles = {
     [filetype in SPAnalysisKnownFiles]: string
 }
 
+const isSPAnalysisFiles = (x: any): x is SPAnalysisFiles => {
+    if (!x) return false
+    if (typeof x !== 'object') return false
+    for (const k of Object.values(SPAnalysisKnownFiles)) {
+        if (typeof x[k] !== 'string') return false
```
WardBrian, line 16: Does this also do the right thing if x does not have one of the keys? I can’t recall if object indexing with an invalid key returns undefined or throws

magland: In that case `x[k]` would be undefined and `typeof x[k] !== 'string'` would be true, so it would return false.

Perhaps we should allow it to be undefined? Certainly in the future we will not want to require data.py, etc. Should we mark files as optional when we get to that point?

WardBrian: If we see a reason to differentiate between “undefined” and “defined but empty” we could do that when the time comes, but I think internally an empty string would be a fine representation for a missing file

gui/src/app/StanSampler/StanSampler.ts
```diff
@@ -12,6 +12,17 @@ export type SamplingOpts = {
...
+export const isSamplingOpts = (x: any): x is SamplingOpts => {
+    if (!x) return false
+    if (typeof x !== 'object') return false
+    if (typeof x.num_chains !== 'number') return false
+    if (typeof x.num_warmup !== 'number') return false
+    if (typeof x.num_samples !== 'number') return false
+    if (typeof x.init_radius !== 'number') return false
+    if (x.seed !== undefined && typeof x.seed !== 'number') return false
+    return true
+}
```
WardBrian, line 24: Should we also verify that e.g. `num_chains` is an integer? 

Not sure what the best way to check this is in JS -- `Math.round(x) === x`?

gui/src/app/StanSampler/StanSampler.ts
```diff
@@ -12,6 +12,17 @@ export type SamplingOpts = {
     seed: number | undefined
 }
 
+export const isSamplingOpts = (x: any): x is SamplingOpts => {
```
WardBrian, line 15: Do you mind also adding a use of this check to SPAnalysisQueryLoading? There is a TODO there for it

WardBrian (commented):
Two small things

gui/src/app/StanSampler/StanSampler.ts
magland: We could check that elsewhere, but I don't think this belongs in the type guard. I believe a type guard should have the sole purpose of guaranteeing that the thing if of the right javascript type.

gui/src/app/StanSampler/StanSampler.ts
magland: Done.

WardBrian (approved):

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
```diff
@@ -96,8 +97,12 @@ export const fetchRemoteAnalysis = async (query: QueryParams) => {
...
         try {
             const parsed = JSON.parse(sampling_opts)
-            // TODO validate the parsed object
-            data.samplingOpts = parsed
+            if (isSamplingOpts(parsed)) {
+                data.samplingOpts = parsed
+            }
+            else {
+                console.error('Invalid sampling_opts in fetchRemoteAnalysis', parsed)
+            }
```
jsoules, line 94: If we're coming this far, it'd probably be good to include a validity check beyond just a type-guard--e.g. natural numbers where appropriate. (I'm okay with the discussion below that suggests it shouldn't be in the type guard proper, but I think if we're going to do the deserialization validity check it's worth it to go the whole way in another fn)

jsoules: I'm imagining most likely a formal parsing function, not in this file but closer to the type definition, that would replace the `JSON.parse()` line above and return either a valid `SamplingOpts` or `undefined`?

jsoules (commented):
I actually did have some suggestions on this, but I see it's merged... I'll go do it separately

PR #94, opened 2024-06-26
prototypes: data.py, data.r, analysis.py, analysis.r

magland:
This is a redo of #74

In addition to being based on the new main branch, the difference is that it uses an unobtrusive method for triggering the prototypes view.

See this comment

```typescript
// We want to be as unobtrusive as possible, so we're not going to hook into the
// routing system for the application. We just have this test for
// prototypes, which is a query parameter that can be set to 1 to enable the
// prototypes. Then below we just render the Prototypes Window on this
// condition.
const query = new URLSearchParams(window.location.search)
const prototypesMode = query.get('prototypes') === '1'
```

So you would do http://localhost:3000?prototypes=1

The idea here is to be able to develop the functionality of generating data.json using data.py or data.r (and also create plots using analysis.py) and then later decide how to integrate it into the UI. It think it would be a mistake to try to do both at the same time.


gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyFileEditor.tsx
```diff
@@ -0,0 +1,159 @@
...
+    width: number
+    height: number
+    outputDiv?: HTMLDivElement | null
+}
+
+const AnalysisPyFileEditor: FunctionComponent<Props> = ({fileName, fileContent, onSaveContent, editedFileContent, setEditedFileContent, readOnly, width, height, outputDiv}) => {
+    const [status, setStatus] = useState<'idle' | 'loading' | 'running' | 'completed' | 'failed'>('idle')
+
+    const loadPyodideInstance = useMemo(() => {
+        let pyodide: PyodideInterface | null = null
```
WardBrian, line 22: Is there any hope of sharing code between this and the data file? Not just for code organization reasons, but I also noticed that this seems to re-download pyodide if you run `data.py` and then switch and run `analysis.py`

gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyPrototype.tsx
```diff
@@ -0,0 +1,93 @@
+import { Splitter } from "@fi-sci/splitter"
+import { FunctionComponent, useEffect, useState } from "react"
+import AnalysisPyFileEditor from "./AnalysisPyFileEditor"
+
+type AnalysisPyPrototypeProps = {
+    width: number
+    height: number
+}
+
+const exampleScript = `import sys
```
WardBrian, line 10: Just to note, the `stanio` package will probably also be helpful here in terms of preparing a useful output object from the draws. You can see what I'm doing in the normal python tinystan here:

https://github.com/WardBrian/tinystan/blob/main/clients/python/tinystan/output.py

WardBrian (commented):
At first look the functionality here looks really good! Have you had a go at plot outputs from webr yet?

A few simplifying suggestions and a question:

magland:
> At first look the functionality here looks really good! Have you had a go at plot outputs from webr yet?

I did look into it, but it's a lot less straightforward than for pyodide (a lot of ugly message passing). I suggest we hold off on that until after we are able to incorporate the analysis.py into the UI.

gui/src/app/prototypes/DataPyPrototype/DataPyFileEditor.tsx
```diff
@@ -0,0 +1,184 @@
...
+            script += '    elif isinstance(x, tuple):\n'
+            script += '        ret = []\n'
+            script += '        for val in x:\n'
+            script += '            ret.append(_sp_serialize(val))\n'
+            script += '        return ret\n'
+            script += '    elif isinstance(x, np.ndarray):\n'
+            script += '        return _sp_serialize(x.tolist())\n'
+            script += '    else:\n'
+            script += '        return x\n'
+            script += 'import json\n'
```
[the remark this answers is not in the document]
magland: This is done.

gui/src/app/prototypes/DataPyPrototype/DataPyFileEditor.tsx
```diff
@@ -0,0 +1,184 @@
...
+    const [status, setStatus] = useState<'idle' | 'loading' | 'running' | 'completed' | 'failed'>('idle')
+    const handleRun = useCallback(async () => {
+        if (status === 'running') {
+            return
+        }
+        if (editedFileContent !== fileContent) {
+            throw new Error('Cannot run edited code')
+        }
+        setStatus('loading')
+        await new Promise(resolve => setTimeout(resolve, 100))
```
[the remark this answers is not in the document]
magland: I have removed this timeout as well as in a couple other places.

There are still a couple timeouts right before pyodid runs, and I have added the following comment to those

```typescript
            // the runPython call is going to be blocking, so we want to give
            // react a chance to update the status in the UI.
```

WardBrian:
> I did look into it, but it's a lot less straightforward than for pyodide (a lot of ugly message passing). I suggest we hold off on that until after we are able to incorporate the analysis.py into the UI.

Sounds good. We may want to look at how the Quarto plugin for webr does things: https://github.com/coatless/quarto-webr/blob/8795c3d75fdc8348734d6c05f1ba6350fb225f8e/_extensions/webr/qwebr-compute-engine.js

gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyFileEditor.tsx
magland: We need to make a decision on whether to use the same pyodide instance for data.py and analysis.py. Since they are going to be depending on a different set of packages (analysis.py will potentially be a lot heavier), I suggest that we use two separate instances even if it means we download twice.

WardBrian:
> Quarto plugin for webr

Here is a very minimal example showing a plot loaded in the same way they do it: https://stackblitz.com/edit/vitejs-vite-6wuedv?file=src%2FApp.tsx

magland:
@WardBrian I have added support for analysis.r based on your hints.

WardBrian:
Nice! Trying it out now

gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyPrototype.tsx
magland: I'm wondering whether we should use a pandas dataframe (long form, all chains in one). We want something that will stand the test of time so that old SP projects won't stop working.

WardBrian: If you want compatibility with existing packages, your best bet will either be something like the tinystan object (which is basically just a dict with ndarray entries), or for maximum interoperability something like ArViz's inferencedata object

magland: I think simpler is better. And we should also try to make it as consistent as possible between R and Python.

WardBrian: R is a slightly different story, but it has a similar sort of lock-in by a package called `posterior`. If whatever we provide is not at least easily convertible to something posterior can understand, it won't be used

magland: One approach could be to provide the data to the script in as raw a format as possible. Then have a standard first few of lines that load it into posterior, or arviz, or pandas, etc. Those conversion lines would be part of the analysis.py/analysis.r.

WardBrian: That seems fine to me

gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyFileEditor.tsx
WardBrian: Having the variables available from the data script (or at the very least, the `data` variable) has some value, I think

WardBrian:
A few notes before this can move to the non-prototype stage:

1. I noticed `yarn build` is raising a bunch of warnings on this branch, possibly related to the pyodide instructions here: https://pyodide.org/en/stable/usage/working-with-bundlers.html#vite

2. Similarly, I think it would make sense to tell vite to separately chunk webr and pyodide outside of index.js

3. Pyodide is currently blocking the main thread. It has web worker support, but it seems to be something you need to do manually: https://pyodide.org/en/stable/usage/webworker.html  -- WebR seems to be in a worker [already/by default](https://docs.r-wasm.org/webr/latest/communication.html).

magland:
> A few notes before this can move to the non-prototype stage:
> 
> 1. I noticed `yarn build` is raising a bunch of warnings on this branch, possibly related to the pyodide instructions here: https://pyodide.org/en/stable/usage/working-with-bundlers.html#vite

If I follow the link in the warning it goes to
https://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility
which explains that the problem is with the pyodide code. I tried following those bundling instructions that you provided, but it didn't help. I could be wrong, but I think we just need to ignore those warnings.

> 2. Similarly, I think it would make sense to tell vite to separately chunk webr and pyodide outside of index.js

I tried lazy importing which resulted in code splitting. However, the split off chunks were just around 16 kb for pyodide and 60 kb for webr... so not worth it. The bulk of the source code seems to be dynamically loaded at runtime.

> 3. Pyodide is currently blocking the main thread. It has web worker support, but it seems to be something you need to do manually: https://pyodide.org/en/stable/usage/webworker.html  -- WebR seems to be in a worker [already/by default](https://docs.r-wasm.org/webr/latest/communication.html).

I agree it would be better to do pyodide in a web worker. But I don't think this should "block" us right now since I wouldn't expect user scripts to be very computationally intensive. You can correct me on that.

WardBrian:
Computationally intensive in terms of number-crunching no, but I certainly think a common use case for `data.py` will be "fetch this existing dataset from the internet and munge it", so freezing during that fetch could be quite annoying

magland:
> Computationally intensive in terms of number-crunching no, but I certainly think a common use case for `data.py` will be "fetch this existing dataset from the internet and munge it", so freezing during that fetch could be quite annoying

If you feel this is important, I can take a crack at it for `data.py`. But for `analysis.py` I think this will prove difficult (if not very difficult) because of the need for this magic:

```
(document as any).pyodideMplTarget = outputDiv;
```

WardBrian:
I think it is at least worth figuring out if it is possible during this prototyping phase. The fact that jupyterlite doesn't freeze its UI means there must be something there, but it definitely seems possible that if we see what the worker solution requires we may decide that the cure is worse than the disease

magland:
> I think it is at least worth figuring out if it is possible during this prototyping phase. The fact that jupyterlite doesn't freeze its UI means there must be something there, but it definitely seems possible that if we see what the worker solution requires we may decide that the cure is worse than the disease

Okay, I'll do it for data.py. (I'm almost positive that jupyterlite use a much more complex system to produce graphical outputs than simply a `document.pyodideMplTarget = div` ... so I'll hold off on the analysis.py part)

WardBrian:
It seems like there are a few different solutions out there, see https://github.com/pyodide/matplotlib-pyodide/issues/6#issuecomment-1242747625

Most of them look like something not too dissimilar to the webR code where you can end up with a list of pictures and separately a list of text outputs from a given program, and then you need to draw them back on the main thread.

magland:
@WardBrian I've got data.py running with a worker.

Do you imagine a need for worker on analysis.py at this point?

WardBrian:
Here's a working (but quite messy) example of pyodide + matplotlib in a worker, in a similar style to how we do webR now: https://gist.github.com/WardBrian/d89e939c43134ea405fd68084e702ddf

I think it is possible that analysis scripts could take a long time to run (in particular, big plots take a long time sometimes), and this also lets us add a 'cancel' button which seems valuable. This should also work for any package which is built on top of matplotlib like seaborne, though it will be worth testing for that later.

magland:
@WardBrian @jsoules wonder if you could try to track down the build error here. I'm stuck.

There seems to be a problem importing loadPyodide from the web worker dataPyWorker.ts

```
[vite:worker] Invalid value "iife" for option "output.format" - UMD and IIFE output formats are not supported for code-splitting builds.
```

WardBrian:
Google immediately suggests adding 

```
  worker: {
    format: 'es',
  },
```
to the vite config, but I'd be at least a little nervous about this causing issues with the other workers we already have. It's also possible to have the worker be a non-module worker and use `importScripts()` for pyodide, but that loses some other niceties

magland:
>   worker: {
>     format: 'es',
>   },
> ```

This seems to have worked! You definitely have a knack.

WardBrian:
> This seems to have worked! You definitely have a knack.

I don't think I'm that far away from having [Tetris Dreams](https://www.scientificamerican.com/article/tetris-dreams/) for the MDN web worker documentation at this point

magland:
Okay I've implemented analysis.py in a worker... and I've also left in the non-worker case for comparison, and in case we want to fall back to that.

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
```diff
@@ -0,0 +1,91 @@
...
+      indexURL: "https://cdn.jsdelivr.net/pyodide/v0.26.1/full",
+      stdout: (x: string) => {
+        self.postMessage({ type: "stdout", data: x });
+      },
+      stderr: (x: string) => {
+        self.postMessage({ type: "stderr", data: x });
+      },
+    });
+    pyodide = p;
+    await pyodide.loadPackage(["numpy", "matplotlib"]);
```
WardBrian, line 16: Passing these as the `packages` argument to the loadPyodide should be faster:

> A list of packages to load as Pyodide is initializing. This is the same as loading the packages with [pyodide.loadPackage()](https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadPackage) after Pyodide is loaded except using the packages option is more efficient because the packages are downloaded while Pyodide bootstraps itself.

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
```diff
@@ -0,0 +1,91 @@
...
+    // here's where we can pass in globals
+    const globals = pyodide.toPy({ _sp_example_global: 5 });
+    const script = MPLPreample + '\n' + code + '\nprint("-----x", len(SP_IMAGES))';
+    let succeeded = false;
+    try {
+      await pyodide.runPython(script, { globals });
+      succeeded = false;
+    }
+    catch (e) {
+      console.error(e);
```
WardBrian, line 78: Should we do something other than just log this? Like post an error message?

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
```diff
@@ -0,0 +1,91 @@
...
+      stdout: (x: string) => {
+        self.postMessage({ type: "stdout", data: x });
+      },
+      stderr: (x: string) => {
+        self.postMessage({ type: "stderr", data: x });
+      },
+    });
+    pyodide = p;
+    await pyodide.loadPackage(["numpy", "matplotlib"]);
+    return pyodide;
```
WardBrian, line 17: Note for later: I was able to get arviz working by adding this code before this return:

```js
    const micropip = pyodide.pyimport("micropip");
    await micropip.install("arviz<0.18");
```


Then this example runs!

```python

import arviz as az

az.style.use("arviz-doc")

data = az.load_arviz_data("non_centered_eight")
az.plot_trace(data, var_names=("tau", "mu"))

plt.show()
```

WardBrian (commented):
Some notes on the worker -- overall a much cleaner implementation than my hacky thing, and I think it's definitely the way to go. Even just playing around with the small example program we have loaded, it's a much smoother experience. 

With an eye toward this becoming a non-prototype, two questions:

1. Should data.py and analysis.py be using the same Worker code? (I lean toward yes, it seems like a reasonable thing to do)
2. Should data.py and analysis.py use the same _Worker instance_? (I lean toward no for an initial implementation)

magland:
> 1. Should data.py and analysis.py be using the same Worker code? (I lean toward yes, it seems like a reasonable thing to do)

I agree. But I think we should make that merge only when the time is right.

> 2. Should data.py and analysis.py use the same _Worker instance_? (I lean toward no for an initial implementation)

I agree. Different dependencies. And I would guess that the user will not usually use both in a single session.

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
```diff
@@ -0,0 +1,91 @@
...
+    matplotlib.pyplot.show = show
+patch_matplotlib(SP_IMAGES)
+`;
+
+const run = async (code: string) => {
+  setStatus("loading");
+  try {
+    const pyodide = await loadPyodideInstance();
+    setStatus("running");
+    // the runPython call is going to be blocking, so we want to give
```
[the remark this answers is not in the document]
magland: removed.

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
```diff
@@ -0,0 +1,91 @@
...
+    // react a chance to update the status in the UI.
+    await new Promise((resolve) => setTimeout(resolve, 100));
+
+    // here's where we can pass in globals
+    const globals = pyodide.toPy({ _sp_example_global: 5 });
+    const script = MPLPreample + '\n' + code + '\nprint("-----x", len(SP_IMAGES))';
```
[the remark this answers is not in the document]
magland: I don't think it hurts to have it here, and in fact it makes things more robust in my mind.

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
```diff
@@ -0,0 +1,91 @@
...
+    let succeeded = false;
+    try {
+      await pyodide.runPython(script, { globals });
+      succeeded = false;
+    }
+    catch (e) {
+      console.error(e);
+    }
+
+    const images = globals.toJs().get("SP_IMAGES");
```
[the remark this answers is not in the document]
magland: Fixed.

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
```diff
@@ -0,0 +1,91 @@
...
+    // the runPython call is going to be blocking, so we want to give
+    // react a chance to update the status in the UI.
+    await new Promise((resolve) => setTimeout(resolve, 100));
+
+    // here's where we can pass in globals
+    const globals = pyodide.toPy({ _sp_example_global: 5 });
+    const script = MPLPreample + '\n' + code + '\nprint("-----x", len(SP_IMAGES))';
+    let succeeded = false;
+    try {
+      await pyodide.runPython(script, { globals });
```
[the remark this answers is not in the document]
magland: Fixed.

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
magland: Done.

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
magland: Very cool. I have implemented this conditionally on whether the script has the word "arviz" in it.

gui/src/app/prototypes/AnalysisPyPrototype/analysisPyWorker.ts
WardBrian: It does look much better!

The images need to be constrained in size in some way I think -- the arviz example I posted below leads to a very large image which scrolls off the screen

WardBrian:
Here's a good proof-of-concept for analysis.R, straight out of bayesplot's examples: https://mc-stan.org/bayesplot/reference/PPC-scatterplots.html#examples

```R
webr::shim_install()
install.packages("bayesplot")

library(bayesplot)
y <- example_y_data()
yrep <- example_yrep_draws()
p1 <- ppc_scatter_avg(y, yrep)
p1


# don't draw line x=y
ppc_scatter_avg(y, yrep, ref_line = FALSE)


p2 <- ppc_scatter(y, yrep[20:23, ], alpha = 0.5, size = 1.5)
p2


# give x and y axes the same limits
lims <- ggplot2::lims(x = c(0, 160), y = c(0, 160))
p1 + lims

p2 + lims


# for ppc_scatter_avg_grouped the default is to allow the facets
# to have different x and y axes
group <- example_group_data()
ppc_scatter_avg_grouped(y, yrep, group)


# let x-axis vary but force y-axis to be the same
ppc_scatter_avg_grouped(y, yrep, group, facet_args = list(scales = "free_x"))
```

This works as is with no modifications to the code. We might want to call `webr::shim_install()` automatically? https://docs.r-wasm.org/webr/latest/packages.html#shimming-install.packages-for-webr

WardBrian:
@magland -- How do we proceed from here? I personally feel like the functionality is good enough at this point that it is worth splitting the PR into two that actually integrate into the rest of the site (assuming you didn't intend for this branch to ever be merged directly -- correct me if I'm wrong there).

I suspect some more things will jump out to us once it is placed in context, and it is probably a good opportunity to think about code organization etc now that it seems we are in a pretty good place in terms of feature exploration. Thoughts?

magland:
> @magland -- How do we proceed from here? I personally feel like the functionality is good enough at this point that it is worth splitting the PR into two that actually integrate into the rest of the site (assuming you didn't intend for this branch to ever be merged directly -- correct me if I'm wrong there).
> 
> I suspect some more things will jump out to us once it is placed in context, and it is probably a good opportunity to think about code organization etc now that it seems we are in a pretty good place in terms of feature exploration. Thoughts?

I didn't have a specific plan. My inclination is to just start new PRs and copy the code over, being sure to reference this discussion in the new PRs.

In terms of code organization, I don't have strong opinions, and I don't think too much thought needs to be put into it. Just tidy things up.

WardBrian:
> My inclination is to just start new PRs and copy the code over, being sure to reference this discussion in the new PRs.

That sounds good to me!

jsoules (commented):
Couple of notes. I didn't get too into the weeds since it seems like we won't be merging this version directly; I'll be more detailed with comments for the subsequent PRs.

magland:
In preparation for incorporating into the UI, this latest commit I did the following.

Combined worker code for data.py and analysis.py so there is now one source file, but still two separate worker instances.

Created pyodidedWorker.ts (worker thread), pyodideWorkerInterface (main thread), and pyodideWorkerTypes (types for messages, status, etc)

Overall I think this implementation is much more solid.

gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyPrototype.tsx
```diff
@@ -0,0 +1,112 @@
...
+          fileName="analysis.py"
+          fileContent={script}
+          onSaveContent={() => setScript(editedScript)}
+          editedFileContent={editedScript}
+          setEditedFileContent={setEditedScript}
+          readOnly={false}
+          outputDiv={outputDiv}
+        />
+      )}
+      <AnalysisPyOutputWindow key={`analysis-py-output-window-${useWorker}`} width={0} height={0} onOutputDiv={setOutputDiv} />
```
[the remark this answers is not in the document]
magland: To me, passing a ref to the div and then utilizing .current on that ref is susceptible to race conditions etc. With the callback, it is much more clear whether the div has been loaded yet, and the state changes on the load. Maybe there's a more standard way to do it, but I have found that this callback mechanism is robust and is more in line with state management that I am familiar with for react.

gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyFileEditorNoWorker.tsx
```diff
@@ -0,0 +1,167 @@
...
+            pre.appendChild(document.createTextNode(x));
+            outputDiv?.appendChild(pre);
+          },
+          packages: ["numpy", "matplotlib"]
+        });
+        pyodide = p;
+        return pyodide;
+      } else {
+        return pyodide;
+      }
```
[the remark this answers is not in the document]
magland: You're right. Fixed.

gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyFileEditor.tsx
```diff
@@ -0,0 +1,174 @@
...
+  >("idle");
+
+  const [analysisPyWorker, setAnalysisPyWorker] = useState<Worker | undefined>(
+    undefined,
+  );
+
+  // worker creation
+  useEffect(() => {
+    const worker = new Worker(analysisPyWorkerURL, {
+      name: "dataPyWorker",
```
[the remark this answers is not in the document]
magland: Fixed.

gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyFileEditor.tsx
```diff
@@ -0,0 +1,174 @@
...
+        const divElement = document.createElement("div");
+        divElement.appendChild(img);
+        outputDiv?.appendChild(divElement);
+      } else if (dd.type === "stdout") {
+        console.log(dd.data);
+        const divElement = document.createElement("div");
+        divElement.style.color = "blue";
+        const preElement = document.createElement("pre");
+        divElement.appendChild(preElement);
+        preElement.textContent = dd.data;
```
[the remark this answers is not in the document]
magland: I suggest this can be cleaned up later.

gui/src/app/prototypes/DataPyPrototype/dataPyWorker.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+const run = async (code: string) => {
+  setStatus("loading");
+  try {
+    const pyodide = await loadPyodideInstance();
+    setStatus("running");
+    // the runPython call is going to be blocking, so we want to give
+    // react a chance to update the status in the UI.
+    await new Promise((resolve) => setTimeout(resolve, 100));
```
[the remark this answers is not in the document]
magland: Removed

gui/src/app/prototypes/DataPyPrototype/dataPyWorker.ts
```diff
@@ -0,0 +1,104 @@
+import { PyodideInterface, loadPyodide } from "pyodide";
+
+let pyodide: PyodideInterface | null = null;
+const loadPyodideInstance = async () => {
```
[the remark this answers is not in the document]
magland: Done

gui/src/app/prototypes/AnalysisPyPrototype/AnalysisPyFileEditor.tsx
```diff
@@ -0,0 +1,174 @@
...
+type Props = {
+  fileName: string;
+  fileContent: string;
+  onSaveContent: () => void;
+  editedFileContent: string;
+  setEditedFileContent: (text: string) => void;
+  readOnly: boolean;
+  width: number;
+  height: number;
+  outputDiv?: HTMLDivElement | null;
```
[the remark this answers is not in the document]
magland: Sorry, a better answer: I think we want to be able to use this component without specifying an output div.

magland:
With this last commit I have separated out the console and image outputs for analysis.py, analysis.r and included console output windows for data.py and data.r

image

WardBrian:
Note: We should probably add a call to https://pyodide.org/en/stable/usage/api/js-api.html#pyodide.loadPackagesFromImports to the pyodide worker to allow users to use other built-in packages that aren't loaded by default. This can actually replace us manually loading numpy, I think, but the special logic is still needed for arviz since that needs to be installed from pip

magland:
@WardBrian should I make a draft of the analysis.R from the main branch and then hand it over to you for creating "draws" object? I didn't want to do this simultaneously if you were planning on working on it.

WardBrian:
Oh, I wasn’t working on it because I thought you wanted to!

I don’t mind either way, if you want to do it that way it’s ready to go

PR #97, opened 2024-06-27
Add functionality to load project files from a gist

### LINKED ISSUE #87: Load from gist
Rewrite #42 on the new main. Need to wait until #84 is merged.

magland:
Fixes #87 

A rewrite of #42

Example:
http://localhost:3000/?project=https://gist.github.com/magland/da3d4143276827609ec9317bb3db8b04

### DIFF
--- a/gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
+++ b/gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
@@ -1,9 +1,13 @@
 import { useSearchParams } from "react-router-dom";
 import { SPAnalysisDataModel, initialDataModel, persistStateToEphemera } from "./SPAnalysisDataModel";
 import { useCallback } from "react";
+import loadFilesFromGist from "./loadFilesFromGist";
+import { loadFromProjectFiles } from "./SPAnalysisReducer";
+import { mapFileContentsToModel } from "./FileMapping";
 
 
 enum QueryParamKeys {
+    Project= "project",
     StanFile = "stan",
     DataFile = "data",
     SamplingOpts = "sampling_opts",
@@ -38,6 +42,7 @@ export const useQueryParams = () => {
     }
 
     const queries: QueryParams = {
+        project: searchParams.get(QueryParamKeys.Project),
         stan: searchParams.get(QueryParamKeys.StanFile),
         data: searchParams.get(QueryParamKeys.DataFile),
         sampling_opts: searchParams.get(QueryParamKeys.SamplingOpts),
@@ -76,8 +81,19 @@ const deepCopy = (obj: any) => {
 }
 
 export const fetchRemoteAnalysis = async (query: QueryParams) => {
-    // any special 'project' query could be loaded here at the top
-    const data: SPAnalysisDataModel = deepCopy(initialDataModel)
+    const projectUri = query.project
+
+    let data: SPAnalysisDataModel = deepCopy(initialDataModel);
+    if (projectUri) {
+        if (projectUri.startsWith('https://gist.github.com/')) {
+            const contentLoadedFromGist = await loadFilesFromGist(projectUri);
+            data = loadFromProjectFiles(data, mapFileContentsToModel(contentLoadedFromGist.files), false);
+            data.meta.title = contentLoadedFromGist.description;
+        } else {
+            // right now we only support loading from a gist
+            console.error('Unsupported project URI', projectUri)
+        }
+    }
 
     const stanFilePromise = query.stan ? tryFetch(query.stan) : Promise.resolve(undefined);
     const dataFilePromise = query.data ? tryFetch(query.data) : Promise.resolve(undefined);
--- a/gui/src/app/SPAnalysis/SPAnalysisReducer.ts
+++ b/gui/src/app/SPAnalysis/SPAnalysisReducer.ts
@@ -109,7 +109,7 @@ const loadFileFromString = (data: SPAnalysisDataModel, field: SPAnalysisKnownFil
     return newData
 }
 
-const loadFromProjectFiles = (data: SPAnalysisDataModel, files: Partial<FieldsContentsMap>, clearExisting: boolean = false): SPAnalysisDataModel => {
+export const loadFromProjectFiles = (data: SPAnalysisDataModel, files: Partial<FieldsContentsMap>, clearExisting: boolean = false): SPAnalysisDataModel => {
     let newData = clearExisting ? initialDataModel : data
     if (Object.keys(files).includes('meta')) {
         newData = loadMetaFromString(newData, files.meta ?? '')
--- a/gui/src/app/SPAnalysis/loadFilesFromGist.ts
+++ b/gui/src/app/SPAnalysis/loadFilesFromGist.ts
@@ -1,0 +1,39 @@
+import { Octokit } from "@octokit/rest"
+
+const loadFilesFromGist = async (gistUri: string): Promise<{ files: { [key: string]: string }, description: string }> => {
+    const parts = gistUri.split('/')
+    const gistId = parts[parts.length - 1]
+    const octokit = new Octokit()
+    const r = await octokit.request('GET /gists/{gist_id}', {
+        gist_id: gistId,
+        headers: {
+            'X-GitHub-Api-Version': '2022-11-28'
+        }
+    })
+    const gist = r.data
+    const description = gist.description || ''
+    const gistFiles = gist.files
+    const files: { [key: string]: string } = {}
+    for (const fname in gistFiles) {
+        const file = gistFiles[fname]
+        if (!file) continue
+        let content = file.content
+        if (content === undefined) continue
+        // gists do not support empty files or whitespace-only files
+        if (content.startsWith('<<empty>>')) {
+            const x = content.slice('<<empty>>'.length)
+            if (x.trim() === '') {
+                content = x
+            }
+        }
+        const fname2 = replaceBarsWithSlashes(fname)
+        files[fname2] = content
+    }
+    return { files, description }
+}
+
+const replaceBarsWithSlashes = (s: string) => {
+    return s.split('|').join('/')
+}
+
+export default loadFilesFromGist
--- a/gui/package.json
+++ b/gui/package.json
@@ -20,6 +20,7 @@
     "@monaco-editor/react": "^4.6.0",
     "@mui/icons-material": "^5.15.17",
     "@mui/material": "^5.15.17",
+    "@octokit/rest": "^21.0.0",
     "jszip": "^3.10.1",
     "monaco-editor": "^0.48.0",
     "plotly.js-cartesian-dist": "^2.33.0",
--- a/gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
+++ b/gui/src/app/SPAnalysis/SPAnalysisContextProvider.tsx
@@ -10,6 +10,7 @@ type SPAnalysisContextType = {
 }
 
 type SPAnalysisContextProviderProps = {
+    //
 }
 
 export const SPAnalysisContext = createContext<SPAnalysisContextType>({
--- a/gui/yarn.lock
+++ b/gui/yarn.lock
@@ -714,6 +714,99 @@
     "@nodelib/fs.scandir" "2.1.5"
     fastq "^1.6.0"
 
+"@octokit/auth-token@^5.0.0":
+  version "5.1.1"
+  resolved "https://registry.yarnpkg.com/@octokit/auth-token/-/auth-token-5.1.1.tgz#3bbfe905111332a17f72d80bd0b51a3e2fa2cf07"
+  integrity sha512-rh3G3wDO8J9wSjfI436JUKzHIxq8NaiL0tVeB2aXmG6p/9859aUOAjA9pmSPNGGZxfwmaJ9ozOJImuNVJdpvbA==
+
+"@octokit/core@^6.1.2":
+  version "6.1.2"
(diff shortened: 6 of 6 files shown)


gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
```diff
@@ -76,8 +81,19 @@ const deepCopy = (obj: any) => {
...
-    // any special 'project' query could be loaded here at the top
-    const data: SPAnalysisDataModel = deepCopy(initialDataModel)
+    const projectUri = query.project
+
+    let data: SPAnalysisDataModel = deepCopy(initialDataModel);
+    if (projectUri) {
+        if (projectUri.startsWith('https://gist.github.com/')) {
+            const contentLoadedFromGist = await loadFilesFromGist(projectUri);
+            data = loadFromProjectFiles(data, mapFileContentsToModel(contentLoadedFromGist.files), false);
+            data.meta.title = contentLoadedFromGist.description;
```
WardBrian, line 91: Should we `return` right after this?

gui/src/app/SPAnalysis/loadFilesFromGist.ts
```diff
@@ -0,0 +1,45 @@
...
+        if (content.startsWith('<<empty>>')) {
+            const x = content.slice('<<empty>>'.length)
+            if (x.trim() === '') {
+                content = x
+            }
+        }
+        // Gists do not support directories This is a placeholder for the future
+        // where we may want a subdirectory structure for project files. Here we
+        // replace the '|' character with '/' to simulate a directory structure.
+        const fname2 = replaceBarsWithSlashes(fname)
```
WardBrian, line 35: Can we keep the code simpler for now by just keeping this trick in our back pocket for later if we need it?

gui/src/app/SPAnalysis/loadFilesFromGist.ts
```diff
@@ -0,0 +1,45 @@
...
+const loadFilesFromGist = async (gistUri: string): Promise<{ files: { [key: string]: string }, description: string }> => {
+    const parts = gistUri.split('/')
+    const gistId = parts[parts.length - 1]
+    const octokit = new Octokit()
+    const r = await octokit.request('GET /gists/{gist_id}', {
+        gist_id: gistId,
+        headers: {
+            'X-GitHub-Api-Version': '2022-11-28'
+        }
+    })
```
WardBrian, line 12: What does this function do if provided a bad gist ID?

gui/src/app/SPAnalysis/loadFilesFromGist.ts
```diff
@@ -0,0 +1,45 @@
...
+        // Gists do not support empty files or whitespace-only files. This
+        // provides a workaround for that by allowing the user to specify an
+        // empty file by starting the content with the string '<<empty>>'. This
+        // is then removed from the content.
+        if (content.startsWith('<<empty>>')) {
+            const x = content.slice('<<empty>>'.length)
+            if (x.trim() === '') {
+                content = x
+            }
+        }
```
WardBrian, line 31: Why would a user want to specify an empty file? Could we instead just have the map this returns be partial?

(in fact, it looks like it could be precisely `Partial<FileRegistry>`?)

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
```diff
@@ -76,8 +81,19 @@ const deepCopy = (obj: any) => {
 }
 
 export const fetchRemoteAnalysis = async (query: QueryParams) => {
-    // any special 'project' query could be loaded here at the top
-    const data: SPAnalysisDataModel = deepCopy(initialDataModel)
+    const projectUri = query.project
+
+    let data: SPAnalysisDataModel = deepCopy(initialDataModel);
+    if (projectUri) {
+        if (projectUri.startsWith('https://gist.github.com/')) {
```
WardBrian, line 88: Is there a reason why we would _need_ to fail on `http://`?

WardBrian (changes requested):
This is nice! It's cool to see that we now have enough infrastructure that a change like this doesn't take much code.

A few concerns:

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
magland: Depends on if we want to support project plus other overrides. If we choose not to support overrides, then we need to provide an alert if they are not being respected. But I would vote for allowing overrides, parameters in addition to project, as it is here.

gui/src/app/SPAnalysis/loadFilesFromGist.ts
magland: I have commented out this code and modified the comment to be:

// In the future, we may want to do the following to support directories
// (gists do not support directories)

gui/src/app/SPAnalysis/loadFilesFromGist.ts
magland: I have commented out this code and modified the comment to be

// In the future, we may want to do the following to support directories
// (gists do not support directories)

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
magland: Is there a reason to support http://?

gui/src/app/SPAnalysis/loadFilesFromGist.ts
magland: It throws an error, which I think is what we want. Error handling on call of fetchRemoteAnalysis could be added.

WardBrian: The rest of the calls in fetchRemoteAnalysis currently catch their own errors (e.g. tryFetch)

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
magland: The URI needs to specify (a) this is a gist and (b) the gist ID. I think it's valuable to require a canonical representation of that information rather than to accept a variety of possible alternatives. We know that we want to support the https:// uri because that's copy/pasted from the gh url box. I don't see a reason to support others... unless you want to do `gist=[id]`... but isn't it best to have only one supported way to do it?

gui/src/app/SPAnalysis/loadFilesFromGist.ts
magland: I guess we should decide what behavior we want in the case where a stan, data, project url is not able to be loaded. Do we just not load that part and print to the console? Or is one error sufficient to fail the entire operation and provide an alert?

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
magland: And in the future, if we want to support a particular revision of a gist, we could support something like

https://gist.github.com/magland/da3d4143276827609ec9317bb3db8b04/822b03367951dc15a4f9be1f35e9af9f2dc050d7

as the canonical representation of that.

gui/src/app/SPAnalysis/loadFilesFromGist.ts
WardBrian: What I did in #39 was populate the text fields with a message, like the stan code would become

`// failed to load from URL_HERE`

But at the moment it is just doing logging

gui/src/app/SPAnalysis/loadFilesFromGist.ts
WardBrian: I really think it is better to not have the code in the file if it is not used. It will always exist in the git history if needed

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
WardBrian: Overrides seem reasonable in this case.

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
WardBrian: I don't think it would hurt if we just accepted anything with `gist.github.com`, but it is not that important

gui/src/app/SPAnalysis/loadFilesFromGist.ts
magland: Removed it.

gui/src/app/SPAnalysis/loadFilesFromGist.ts
magland: Removed the commented code.

WardBrian (approved):
Looks good!

I'm going to open another PR after this is merged to look a bit at indicating to the user if loading failed, so we don't need to address here

gui/src/app/SPAnalysis/loadFilesFromGist.ts
magland: I have implemented fallback text like this

```
const stanFilePromise = query.stan ? tryFetch(query.stan, `// Unable to load from ${query.stan}`) : Promise.resolve(undefined);
    const dataFilePromise = query.data ? tryFetch(query.data, `Unable to load from ${query.data}`) : Promise.resolve(undefined);
    const sampling_optsPromise = query.sampling_opts ? tryFetch(query.sampling_opts, undefined) : Promise.resolve(undefined);
```

and I think an exception is appropriate when the entire project does not load.

Does that sound okay?

WardBrian: I would prefer something like 

```
    const stanFilePromise = query.stan ? tryFetch(query.stan) : Promise.resolve("");
```

and then have an `else` on the `if (stanFileContent)` following the await. I don't think it is necessary to hand the fallback value to tryFetch just to get it back ourselves.

If you want I can also do this separately

magland: Okay I'll let you work on that. But a note... there are two reasons why the stanFileContent might be empty... only one should result in the message in the editor.

WardBrian: Yeah, I think the `if` would become an `if (stanFileContent !== undefined)`, with the else then only firing when tryfetch returned undefined.

If you're not interested in doing that here, this is good to merge as far as I am concerned!

magland:
@jsoules I'll let you merge once you give it the okay.

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
```diff
@@ -65,20 +70,36 @@ const deepCopy = (obj: any) => {
...
+            const contentLoadedFromGist = await loadFilesFromGist(projectUri);
```
WardBrian, line 78: Even if we don't do anything in the `catch` for now, we should probably wrap this call in a try - right now it just logs an error to the console and prevents the rest of the hook that calls it from running

magland: Do we want to allow it to continue and load the other parts of the query (if they are present) even if the main project load failed? I was thinking we could catch the error and put an alert popup that project couldn't load and then return.

WardBrian: I don't feel strongly either way, if other things are provided and they succeed that seems fine to me. An alert seems like the right thing to do, but anything is probably better than propagating the exception

jsoules (commented):
I'm okay with this as stands, though I do think we may want to move a couple functions out of the reducer.

Agree with @WardBrian's request for a try-catch in gist file loading, though the exact error handling strategy can be deferred to a later PR.

gui/src/app/SPAnalysis/SPAnalysisReducer.ts
```diff
@@ -102,7 +102,7 @@ const loadFileFromString = (data: SPAnalysisDataModel, field: SPAnalysisKnownFil
     return newData
 }
 
-const loadFromProjectFiles = (data: SPAnalysisDataModel, files: Partial<FieldsContentsMap>, clearExisting: boolean = false): SPAnalysisDataModel => {
+export const loadFromProjectFiles = (data: SPAnalysisDataModel, files: Partial<FieldsContentsMap>, clearExisting: boolean = false): SPAnalysisDataModel => {
```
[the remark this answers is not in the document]
magland: Happy to move it. Where should it go @jsoules

gui/src/app/SPAnalysis/SPAnalysisQueryLoading.ts
magland: There's now a try/catch and an alert message; the rest of the loading skipped. I feel this is good for now.

magland:
@WardBrian shall we merge this as is, so that you can do the formatting and we can move forward? We can move that function above afterwards.

gui/src/app/SPAnalysis/SPAnalysisReducer.ts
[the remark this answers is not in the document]
jsoules: Sorry, missed this. I would say it should probably go in `SPAnalysisSerialization` but open to other suggestions.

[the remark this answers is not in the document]
WardBrian: I like it in the Serialization file.

@magland want to move it there, fix the merge conflict that has shown up in the QueryLoading file and merge?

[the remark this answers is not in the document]
magland: Sounds good. I made the move.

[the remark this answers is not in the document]
magland: fixing conflicts now

Issue #23, opened 2024-05-17
Consolidate inline styling to CSS classes.


jsoules:
Additionally, reduce Javascript-level layout computations where possible, in favor of CSS computations or deferring to the browser.

PR #99, opened 2024-06-27
minor: fix editor toolbar styling

### LINKED ISSUE #98: fix styling issue on upper-right corner of editor
image

magland:
Fixes #98 

@WardBrian @jsoules this is very minor, either of you can go ahead and merge.


WardBrian (approved):
Confirmed it fixes that visual artifact. Do we already have an issue about moving away from inline styling? I'll make one now

WardBrian:
Oh we do, #23

PR #100, opened 2024-06-27
implement error/warning squiggles in stan editor

### LINKED ISSUE #76: Improve integration into monaco editor
Besides the functionality in #73, it would also be cool to generate the traditional red squiggles underneath an error. 

Doing so should be just extracting the line, column, and message from a warning/error and then calling [`setModelMarkers`](https://microsoft.github.io/monaco-editor/typedoc/functions/editor.setModelMarkers.html) but I suspect doing this in a way React is happy with may be tricky.

magland:
Fixes #76

### DIFF
--- a/gui/src/app/FileEditor/StanFileEditor.tsx
+++ b/gui/src/app/FileEditor/StanFileEditor.tsx
@@ -240,18 +240,51 @@ const stancErrorsToCodeMarkers = (stancErrors: StancErrors) => {
         }
 
         if ((lineNumber !== undefined) && (startColumn !== undefined) && (endColumn !== undefined)) {
+            const isWarning = x.toLowerCase().startsWith('warning')
             cm.push({
                 startLineNumber: lineNumber,
                 startColumn: startColumn + 1,
                 endLineNumber: lineNumber,
                 endColumn: endColumn + 1,
-                message: x,
-                severity: x.toLowerCase().startsWith('warning') ? 'warning' : 'error'
+                message: isWarning ? getWarningMessage(x) : getErrorMessage(x),
+                severity: isWarning ? 'warning' : 'error'
             })
         }
     }
     return cm
 }
 
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Adapted from https://github.com/WardBrian/vscode-stan-extension
+function getWarningMessage(message: string) {
+    let warning = message.replace(/Warning.*column \d+: /s, "");
+    warning = warning.replace(/\s+/gs, " ");
+    warning = warning.trim();
+    warning = message.includes("included from")
+      ? "Warning in included file:\n" + warning
+      : warning;
+    return warning;
+}
+
+function getErrorMessage(message: string) {
+    let error = message;
+    // cut off code snippet for display
+    if (message.includes("------\n")) {
+      error = error.split("------\n")[2];
+    }
+    error = error.trim();
+    error = message.includes("included from")
+      ? "Error in included file:\n" + error
+      : error;
+
+    // only relevant to vscode-stan-extension:
+    // error = error.includes("given information about")
+    //   ? error +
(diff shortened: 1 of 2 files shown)


WardBrian:
I think it makes sense to do some munging on the error messages so they look nicer on the hover-over, e.g. removing the 'preview' stanc gives you. See:

https://github.com/WardBrian/vscode-stan-extension/blob/main/src/linter.ts#L38
https://github.com/WardBrian/vscode-stan-extension/blob/main/src/linter.ts#L48

Additionally, with this implemented, do we still see value in the "pop-up" error window?

magland:
> Additionally, with this implemented, do we still see value in the "pop-up" error window?

Not sure about that. But can we keep it in there for the time being? Not ready to say goodbye :)

WardBrian:
Sure, no need to delete it in the same PR -- just speculating for the future!

magland:
> I think it makes sense to do some munging on the error messages so they look nicer on the hover-over, e.g. removing the 'preview' stanc gives you. See:
> 
> https://github.com/WardBrian/vscode-stan-extension/blob/main/src/linter.ts#L38 https://github.com/WardBrian/vscode-stan-extension/blob/main/src/linter.ts#L48

I was about to use that code, but saw that it's GPL. Do I have your permission?

WardBrian:
Ah, yes. The project I forked from is GPL, but I am the sole author of that file, and I happily grant permission.

magland:
Okay, the messages are now reformatted and they look nice!

Another comment about the syntax window... I think it's nice to not just rely on seeing a slight yellow squiggle... so even if we do remove that window, we'll want to be able to display something prominently that there are issues (and at that point, I'd like to click on that notification and get the details)

gui/src/app/FileEditor/StanFileEditor.tsx
```diff
@@ -213,5 +215,76 @@ const stringChecksum = (str: string) => {
...
+            }
+            else if ((section.startsWith('column ')) && (startColumn === undefined)) {
+                const cols = section.slice('column '.length).split(' to ')
+                startColumn = parseInt(cols[0])
+                endColumn = cols.length > 1 ? parseInt(cols[1].slice('column '.length)) : startColumn + 1
+            }
+        }
+
+        if ((lineNumber !== undefined) && (startColumn !== undefined) && (endColumn !== undefined)) {
+            const isWarning = x.toLowerCase().startsWith('warning')
```
WardBrian, line 243: Thoughts on having a loop over warnings and a loop over errors to avoid needing to back this information out again? I think if the above logic to find the start/end was its own function the two loops wouldn't need to share too much duplicate code

gui/src/app/FileEditor/TextEditor.tsx
```diff
@@ -60,10 +71,35 @@ const TextEditor: FunctionComponent<Props> = ({defaultText, text, onSaveText, ed
...
+        const model = editor.getModel()
+        if (model === null) return
+        const toSeverity = (s: 'error' | 'warning' | 'hint' | 'info'): monaco.MarkerSeverity => {
+            switch (s) {
+                case 'error': return monaco.MarkerSeverity.Error
+                case 'warning': return monaco.MarkerSeverity.Warning
+                case 'hint': return monaco.MarkerSeverity.Hint
+                case 'info': return monaco.MarkerSeverity.Info
+            }
+        }
```
WardBrian, line 88: Mind moving this helper function out of the hook?

WardBrian (commented):
Two style comments but otherwise I think this looks great!

gui/src/app/FileEditor/StanFileEditor.tsx
magland: Done

gui/src/app/FileEditor/TextEditor.tsx
magland: Done

PR #101, opened 2024-06-27
[WIP] Export project to Gist

magland:
Adds functionality for saving a project to Gist. This includes #97 content, so that should get merged first.

To try it out, click on "Export" and then there's a button "Export to GitHub Gist". It will walk you through the steps and provide instructions on providing an access token with proper scope.

In a separate PR we can figure out what to do about updating an existing gist.

To summarize:
* Export
* Export to GitHub Gist
* Read instructions on getting the PAT
* Enter the PAT
* Click "Export to GitHub Gist"
* Retrieve the url to the gist AND the shareable SP url that loads the gist

### DIFF
--- a/gui/src/app/pages/HomePage/ExportWindow.tsx
+++ b/gui/src/app/pages/HomePage/ExportWindow.tsx
@@ -1,9 +1,10 @@
-import { FunctionComponent, useContext } from "react"
+import { FunctionComponent, useCallback, useContext, useState } from "react"
 
-import { mapModelToFileManifest } from "../../SPAnalysis/FileMapping"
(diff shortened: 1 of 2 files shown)


magland:
I think there should be an distinction between "export" and "save". Export to a zip, save to a gist. So I'm going to make this two separate buttons on the left panel and we'll see how this looks.

WardBrian:
> I think there should be an distinction between "export" and "save". Export to a zip, save to a gist. So I'm going to make this two separate buttons on the left panel and we'll see how this looks.

I think I agree, but I would personally swap the names, with save meaning to the local file system, and export meaning to another service (gist)

magland:
Replaced by #114

PR #102, opened 2024-06-28
rearrange main layout

magland:
Moved data editor to left panel, underneath the stan editor, to make more room for the sampling controls and sampling output panel on the right. Made some other minor adjustments of spacing and sizing. Simplified the data flow in the HomePage. Replaces #45

image

### DIFF
--- a/gui/src/app/pages/HomePage/HomePage.tsx
+++ b/gui/src/app/pages/HomePage/HomePage.tsx
@@ -30,10 +30,7 @@ const HomePage: FunctionComponent<Props> = ({ width, height }) => {
 }
 
 const HomePageChild: FunctionComponent<Props> = ({ width, height }) => {
-    const { data, update } = useContext(SPAnalysisContext)
-    const setSamplingOpts = useCallback((opts: SamplingOpts) => {
-        update({ type: 'setSamplingOpts', opts })
-    }, [update])
+    const { data } = useContext(SPAnalysisContext)
 
     const [compiledMainJsUrl, setCompiledMainJsUrl] = useState<string>('')
 
@@ -54,7 +51,7 @@ const HomePageChild: FunctionComponent<Props> = ({ width, height }) => {
         }
     }, [width])
 
-    const topBarHeight = 25
+    const topBarHeight = 22
 
     useEffect(() => {
         document.title = "Stan Playground - " + data.meta.title;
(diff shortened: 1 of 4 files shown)


WardBrian (approved):

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -128,79 +111,95 @@ const determineShouldBeInitiallyCollapsed = (width: number) => {
...
-    dataFileContent: string
-    dataIsSaved: boolean
-    samplingOpts: SamplingOpts
-    setSamplingOpts: (opts: SamplingOpts) => void
 }
 
-const LowerRightView: FunctionComponent<LowerRightViewProps> = ({ width, height, compiledMainJsUrl, dataFileContent, dataIsSaved, samplingOpts, setSamplingOpts }) => {
+const LowerRightView: FunctionComponent<LowerRightViewProps> = ({ width, height, compiledMainJsUrl }) => {
+    const { data, update } = useContext(SPAnalysisContext)
+    const dataIsSaved = data.dataFileContent === data.ephemera.dataFileContent
```
jsoules, line 179: I see this as a definition internal to the data model, & think it'd probably be better to keep this internal to that file & expose the check via a function (though that function would probably just be `const dataIsSaved = (data: SPAnalysisDataModel) => (data.dataFileContent === data.ephemera.dataFileContent)`)

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -128,79 +111,95 @@ const determineShouldBeInitiallyCollapsed = (width: number) => {
...
+
+const RightView: FunctionComponent<RightViewProps> = ({ width, height, compiledMainJsUrl }) => {
+    return (
+        <LowerRightView
+            width={width}
+            height={height}
+            compiledMainJsUrl={compiledMainJsUrl}
+        />
+    )
+}
```
jsoules, line 169: Seems a little weird to have a `RightView` component whose sole content is the `LowerRightView` component. Are we planning to expand this in the future?

jsoules (commented):
A couple minor things, but we can also just revisit later.

gui/src/app/pages/HomePage/HomePage.tsx
magland: Oops that was a remnant of the prior layout. Removed the extra component.

gui/src/app/pages/HomePage/HomePage.tsx
magland: I have added the function and have named it modelHasUnsavedDataFileChanges to go along with the existing modelHasUnsavedChanges

PR #103, opened 2024-06-28
Show error in text boxes if fetch fails

WardBrian:
This was a feature in #39 I forgot to port forward

Currently nothing is displayed if loading the sampling options fails


magland:
I guess this will conflict with #97. Should I modify #97 to use this? Or what should we do?

WardBrian:
If you want to pull this into #97 that's good with me

magland:
> If you want to pull this into #97 that's good with me

Done. Closing.

PR #108, opened 2024-06-28
minor: Remove unnecessary text param in onSave functions

magland:
With the new data model structure, the text being returned in onSaveContent is no longer used. So I've removed that parameter.


WardBrian (approved):
Looks good to me

PR #110, opened 2024-06-28
Use release 2021.3.0 of TBB

WardBrian:
The patch we need is now officially in a released version, so we don't have to use a specific commit any longer.


jsoules:
Rebuilding the docker image works fine and the result successfully compiles the demo code. I think this is good to go, but I'll note that I do get a warning about mixing the case between a `FROM` and an `AS` in the docker files somewhere.

I don't see that in the files touched by this PR though, so I think you're good to go.

jsoules (approved):
Tch, missed it right under my own eyes. Good call.

Issue #107, opened 2024-06-28
Formatter and linter for the frontend


WardBrian:
I think at least a few of the recent merge conflicts we've had on PRs have been due to formatting differences between editors rather than actual content changing. It would be nice to have a consistent style we can all just tell our editor to use and then forget about it.

I don't have a super strong opinion on which, but to make this concrete I would suggest [Prettier](https://prettier.io/) with a minimal config. For Stan-Web-Demo I used
```json
{
  "tabWidth": 2,
  "semi": true,
  "arrowParens": "avoid"
}
```

magland:
Sounds good. Shall we try to get to a place where we have no outstanding PRs and then sweep through the code and (a) change Analysis to Project (b) enforce prettier formatting?

WardBrian:
I think it would be beneficial to do it earlier (at least before the webr/pyodide stuff gets merged)

Right now is probably the fewest outstanding PRs we’ve had in a while, especially if we merge #102 and #97 today, so I would propose doing it ~now

magland:
Cool. Once those two are merged (from my perspective they are ready to go), could you make a utility script that formats all .ts/.tsx files, and then we can review whether that is suitable? Do we want to have a pre commit hook? CI check? VSCode setting?

WardBrian:
We already have pre-commit on the backend — that seems like a good choice locally, but setting up your VSCode can also help obviously. 

I will update our config and run the full format after they’re merged

magland:
Sounds good... I snuck one more in there #108 which I think could be merged right away.

Issue #64, opened 2024-06-13
rename "analysis" to "project" throughout


magland:
Let's wait until things stabilize more and then we can make this change.

See #54

PR #111, opened 2024-06-28
Add prettier config, format existing files

WardBrian:
The 3 commits in this PR:

- Add config for prettier, both in the pre-commit-config and as a script in our package.json
- Run prettier on the project (this is the majority of the diff), closing #107
- Replace usages of the term "Analysis" with "Project", closing #64


jsoules (approved):
Checked the first and last commits. The pre-commit setup looks reasonable, and the replacement of SPAnalysis -> Project seems to have gone smoothly. Confirmed that there are no appearances of "analysis" in the code base anymore, other than in the stanc javascript (which is referring to something else).

Ran the gui locally with the changes and was still able to compile and sample from a model, so nothing seems obviously broken. I think we're good to merge.

At @WardBrian 's suggestion I also installed the Prettier VSCode extension and enabled it for this workspace.

jsoules:
Oh, should the config exclude running against CSS files? Looks like those got updated too--not sure if this was intentional.

WardBrian:
I think all it really does in CSS files is unify the indentation, which I’m happy with

PR #114, opened 2024-06-29
save project as gist

magland:
Replaces #101 because needed to rebase on the prettier formatted main branch.

Renamed import / export to "Load project" / "Save project". Try it out and see how it feels.


gui/src/app/pages/HomePage/SaveProjectWindow.tsx
```diff
@@ -0,0 +1,208 @@
...
+  return (
+    <div style={{ maxWidth: 800 }}>
+      <h3>Save to GitHub Gist</h3>
+      <p>
+        In order to save this project as a GitHub Gist, you will need to provide
+        a GitHub Personal Access Token.&nbsp; This token will be used to
+        authenticate with GitHub and create a new Gist with the files in this
+        project.&nbsp; You can create a new Personal Access Token by visiting
+        your{" "}
+        <a href="https://github.com/settings" target="_blank" rel="noreferrer">
```
WardBrian, line 135: I think we can link directly through to https://github.com/settings/tokens, no?

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
```diff
@@ -0,0 +1,208 @@
...
+          GitHub settings
+        </a>
+        .&nbsp; Go to <i>Developer settings</i> and <i>Tokens (classic)</i>
```
WardBrian, line 138: I just checked, this _does_ also work with the beta fine-grained tokens. You just need to give it the gist permission and nothing else.

I think we probably should prefer those, since it seems like that is the direction github is heading?

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
```diff
@@ -0,0 +1,208 @@
...
+        .&nbsp; Generate a new classic token and be sure to only grant gist
+        scope with an expiration date.&nbsp; Copy the token and paste it into
+        the field below.
+      </p>
+      <p>
+        For security reasons, your token will not be saved in this
+        application,&nbsp; so you may want to store it securely in a text file
+        for future use.
```
WardBrian, line 146: I thought we had decided local storage was probably fine for this?

gui/src/app/pages/HomePage/saveAsGitHubGist.ts
WardBrian, line 1: Thoughts on moving this code either closer to or into the same file as the gist loading code? Right now that is in `./Project` rather than in `./Pages/HomePage`, which feels like a better place

WardBrian (commented):
Finally got around to trying this out - looking pretty good to me! A few comments mostly on the prose of the instructions (we may also want to include a screen shot or gif?)

gui/src/app/pages/HomePage/saveAsGitHubGist.ts
```diff
@@ -0,0 +1,74 @@
+import { Octokit } from "@octokit/core";
+
+const saveAsGitHubGist = async (
+  files: { [key: string]: string },
+  o: { defaultDescription: string; personalAccessToken: string },
+) => {
+  const { defaultDescription, personalAccessToken } = o;
+  const description = prompt(
```
WardBrian, line 8: Oh also -- this prompt is redundant with the title box already on the save page, right?

I guess it's nice as an "are you sure"?

magland: This was intentional. When saving to a gist it's important to specify an appropriate title/description, and I often find myself forgetting to do this, even when the option is there in the table, and so this is a very helpful opportunity.

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
magland: Done. In fact it now goes directly to fine-grained tokens.

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
magland: Sounds good. I missed the "Account Permissions" section. Yeah that makes sense.

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
magland: I think it is okay, but if it's okay, let's make that a future feature. I want to think through security implications more carefully and how to present this to the user. I think saving in browser should be opt-in only.

gui/src/app/pages/HomePage/saveAsGitHubGist.ts
magland: I created a new folder called gists that now has both files.

magland:
Unless you want to discuss the storing PAT to browser further, this might be ready to go.

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
```diff
@@ -0,0 +1,213 @@
+import { FunctionComponent, useCallback, useContext, useState } from "react";
```
jsoules, line 1: Just noting that this file is getting a little chunky--might consider splitting into separate files for the major components at some point.

jsoules (approved):
Couple cleanup suggestions for the main code file, but otherwise I think this is in good shape.

gui/src/app/gists/saveAsGitHubGist.ts
```diff
@@ -0,0 +1,74 @@
...
+    "SAVING AS PUBLIC GIST: Enter a description:",
+    defaultDescription,
+  );
+  if (!description) {
+    throw Error("No description provided");
+  }
+  const octokit = new Octokit({
+    auth: personalAccessToken,
+  });
+  const files2: { [key: string]: { content: string } } = {};
```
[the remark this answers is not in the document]
magland: Done

gui/src/app/gists/saveAsGitHubGist.ts
```diff
@@ -0,0 +1,74 @@
+import { Octokit } from "@octokit/core";
+
+const saveAsGitHubGist = async (
+  files: { [key: string]: string },
+  o: { defaultDescription: string; personalAccessToken: string },
```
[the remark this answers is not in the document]
magland: Done

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
magland: agreed

jsoules:
Changes look good. I think we're ready to merge if @WardBrian is okay with the language.

WardBrian (approved):

PR #115, opened 2024-07-01
auto deploy main branch to vercel

magland:
This will auto deploy main branch to https://stan-playground.vercel.app

If we decide to go with stan-playground.flatironinstitute.org, this will effectively deploy to that domain.


magland:
Too many mistake commits. Closing. Will open new PR.

PR #116, opened 2024-07-01
Add Vercel deployment workflow

magland:
Auto deploy main branch to stan-playground.vercel.app

If we decide to go with stan-playground.flatironinstitute.org it will effectively deploy to that domain as well.

### DIFF
--- a/.github/workflows/vercel_deploy.yml
+++ b/.github/workflows/vercel_deploy.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-      - vercel-deploy  # for initial test
+      - vercel-deploy-2  # for initial test
   # manual trigger
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


WardBrian:
Looks good. If you wanted to move it into the other .yaml file you could have the deploy gated on the fact that the tests pass, which might be nice, but I think this is fine as is

.github/workflows/vercel_deploy.yml
```diff
@@ -0,0 +1,36 @@
+name: Vercel deploy
+
+on:
+  push:
+    branches:
+      - main
+      - vercel-deploy-2  # for initial test
+  # manual trigger
+  workflow_dispatch:
```
jsoules, line 9: Sorry, does this mean this task *is* triggered manually or is that a way *to* trigger it manually?

jsoules (approved):
Seems reasonable enough. I haven't tried executing it as I don't know that I actually have keys to do so, but I'm satisfied if you have done so.

.github/workflows/vercel_deploy.yml
magland: It can optionally be triggered manually in addition to automatically.

PR #117, opened 2024-07-01
Move linting code to under Stanc, tests

### LINKED ISSUE #105: Encapsulate string/error-parsing logic from StanFileEditor.tsx
A lot of the logic for parsing error messages from stanc is currently located in the editor front-end component. I believe it would be more robust to put it in a separate file or closer to the stanc interface.

WardBrian:
Closes #105


gui/test/app/Stanc/Linting.test.ts
```diff
@@ -0,0 +1,112 @@
...
+Ill-typed arguments supplied to assignment operator =:
+The left hand side has type
+  array[] int
+and the right hand side has type
+  int
+`;
+
+describe("Linting", () => {
+  describe("position detection", () => {
+    test("random message returns undefined", () => {
```
jsoules, line 43: Consider adding a case where message is empty; the `if (!message)` line doesn't seem to be tested.

gui/src/app/Stanc/Linting.ts
```diff
@@ -0,0 +1,117 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// Adapted from https://github.com/WardBrian/vscode-stan-extension
+
+const locationFromMessage = (message: string): Position | undefined => {
+  if (!message) return;
+  // format is "in 'filename', line (#)), column (#) to (line #,)? column (#)"
+  const start = message.matchAll(/'.*', line (\d+), column (\d+)( to)?/g);
+  if (!start) {
+    return undefined;
+  }
```
jsoules, line 61: This shouldn't be a possible case: if there are no matches it'll return an empty iterator (per the documentation) & the code coverage report says this condition isn't checked.

gui/test/app/Stanc/Linting.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    test("should detect the position of a single line, single column message", () => {
+      const position = locationFromMessage(jacobianWarning);
+      expect(position).toBeDefined();
+      assert(position);
+      expect(position.startLineNumber).toEqual(position.endLineNumber);
+      expect(position.startColumn).toEqual(position.endColumn);
+
+      expect(position.startLineNumber).toEqual(1);
+      expect(position.startColumn).toEqual(13); // 0 vs 1 based indexing
```
jsoules, line 57: Which one is which? Lines are 1-based and columns are 0-based?

gui/test/app/Stanc/Linting.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    });
+
+    test("should detect the position of a single line, multi-column error", () => {
+      const position = locationFromMessage(multiColumnError);
+      expect(position).toBeDefined();
+      assert(position);
```
jsoules, line 63: ?
What's the case where these would both be needed?

gui/test/app/Stanc/Linting.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      expect(position.startLineNumber).toEqual(3);
+      expect(position.startColumn).toEqual(5);
+      expect(position.endColumn).toEqual(11);
+    });
+
+    test("should detect the position of a multi-line, multi-column error", () => {
+      const position = locationFromMessage(multiLineError);
+      expect(position).toBeDefined();
+      assert(position);
```
jsoules, line 75: (ditto)

jsoules (approved):
I'm fine with this as is, but spent way too much time making fussy little recommendations.

Thanks for including the tests; they're great for clarifying things. Looking at the coverage reports, you may want to include a test for both warnings and errors from included files.

You may also wish to write a test for `stancErrorsToCodeMarkers`, it's pretty trivial but I noticed it was currently uncovered, and it would probably cover `stancMessageToMarker` by incorporation. I suppose the behaviors to confirm would be that there are no empty elements in the list, that the errors are all mapped before the warnings, and that the messages track with the severity or something?

gui/src/app/Stanc/Linting.ts
```diff
@@ -0,0 +1,117 @@
...
+  if (message.includes("------\n")) {
+    error = error.split("------\n")[2];
+  }
+  error = error.trim();
+  error = message.includes("included from")
+    ? "Error in included file:\n" + error
+    : error;
+
+  return error;
+}
```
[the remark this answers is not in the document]
WardBrian: Not really. We could combine some of it, but one of the main differences is whitespace in warnings is, in general, not meaningful, but in errors it often is.

gui/src/app/Stanc/Linting.ts
```diff
@@ -0,0 +1,117 @@
...
+  let endColumn = startColumn;
+
+  if (lastMatch[3]) {
+    // " to" was matched
+    const end = message.match(/to (line (\d+), )?column (\d+)/);
+    if (end) {
+      if (end[1]) {
+        endLineNumber = parseInt(end[2]);
+      }
+      endColumn = parseInt(end[3]) + 1;
```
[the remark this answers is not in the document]
WardBrian: Yes

```node
> "foobarbaz".match(/(foo(bar))?(baz)/)
[
  'foobarbaz',
  'foobar',
  'bar',
  'baz',
  index: 0,
  input: 'foobarbaz',
  groups: undefined
]
> "snoobarbaz".match(/(foo(bar))?(baz)/)
[
  'baz',
  undefined,
  undefined,
  'baz',
  index: 7,
  input: 'snoobarbaz',
  groups: undefined
]
```

gui/test/app/Stanc/Linting.test.ts
WardBrian: Stanc emits 1-based line numbers and 0-based column numbers (I know.).  Monaco expects 1-based for both:

https://microsoft.github.io/monaco-editor/typedoc/classes/Position.html#lineNumber
https://microsoft.github.io/monaco-editor/typedoc/classes/Position.html#column

gui/src/app/Stanc/Linting.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  return { startLineNumber, startColumn, endLineNumber, endColumn };
+};
+
+function getWarningMessage(message: string) {
```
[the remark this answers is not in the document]
WardBrian: Both of these were a result of this code coming from https://github.com/WardBrian/vscode-stan-extension, but since we're already making changes to it we might as well make this one I suppose

gui/test/app/Stanc/Linting.test.ts
jsoules: nah, that's fine if it's just a linter thing. I just wasn't sure if there was an actual issue at test.

WardBrian:
I added the tests you requested, expect for included files -- since we don't currently support that, I just removed the code. A bunch of stuff would need to change first anyway ;)

gui/src/app/Stanc/Linting.ts
```diff
@@ -0,0 +1,103 @@
...
+  const codeMarkers = [
+    ...(stancErrors.errors || []).map((error) =>
+      stancMessageToMarker(error, "error"),
+    ),
+    ...(stancErrors.warnings || []).map((warning) =>
+      stancMessageToMarker(warning, "warning"),
+    ),
+  ];
+
+  return codeMarkers.filter((marker) => marker !== undefined) as CodeMarker[];
```
jsoules, line 26: This is fine, no change requested. But I wonder whether it's even possible to generate one of these? Or was this what the linter was complaining about

jsoules: Apologies, that was a bit gnomic. I mean is it possible for this list to even have any undefined elements? Or is the filter just to keep TSC happy

jsoules: Per our conversation, it is possible to generate warnings with no line numbers (e.g. an entirely whitespace stan program) which would cause `undefined`s to be legitimately returned, so this check may be necessary in some edge cases.

PR #120, opened 2024-07-03
Python analysis window

magland:
Merged code from #94 

Created "Analysis (Py)" tab.

Next step: provide sampling draws as input to script


magland:
You can now run the following in analysis.py

```python
import pandas as pd

df = pd.DataFrame(sp_sampling.draws, columns=sp_sampling.parameter_names)

print(df)
```

magland:
Here's that example as gist project

http://localhost:3000/?project=https://gist.github.com/magland/8d987a04c9db7bee8636960a69ae7c7f

magland:
There are a number of things to think about. What should the global variable (now sp_sampling) be called? Right now sp_sampling.draws has all the chains concatenated, and you also get sp_sampling.num_chains. It reflects the internal structure of the draws in the data model. Should that be adjusted? What's most intuitive for the user creating analysis.py?

WardBrian:
Right now the UI lets you click over to analysis.py/analysis.R when you have not yet run sampling. Moving those tabs down to the area we only show after sampling has completed seems like a good idea to prevent this

WardBrian:
> Should that be adjusted? What's most intuitive for the user creating analysis.py?

The most intuitive will depend pretty strongly on the user and their use case, but I think there are 3 main 'types' of uses:

- the user cares about the markov chains as markov chains, so having a `num_samples` x `num_chains` x `num_parameters` (or the transpose of this) numpy array will be the most useful for them
- The user is using an existing tool which relies on pandas dataframes or arviz InferenceData, so whatever we provide should be easy to convert to these
- The user cares about one specific parameter in the model, in which case providing a way to extract that one item as a numpy array of the correct shape incredibly useful. This is what cmdstanpy calls [`stan_variable`](https://mc-stan.org/cmdstanpy/api.html#cmdstanpy.CmdStanMCMC.stan_variable) and is implemented in `stanio`. Note that the reason this is tricky is when I say "parameter", I am including things like a parameter declared to be a matrix, not just one column of output

gui/src/app/pyodide/pyodideWorker/pyodideWorkerTypes.ts
```diff
@@ -0,0 +1,69 @@
...
+export type PydodideWorkerStatus =
+  | "idle"
+  | "loading"
+  | "running"
+  | "completed"
+  | "failed";
+
+export const isPydodideWorkerStatus = (x: any): x is PydodideWorkerStatus => {
+  return ["idle", "loading", "running", "completed", "failed"].includes(x);
+};
```
jsoules, line 69: It has its own drawbacks, but defining the status as an enum allows you to pull the keys/values of the enum as a list, which might make the `includes` check here more intuitive. (That said we probably aren't expanding the number of statuses any time soon, so maybe not worth worrying about?)

gui/src/app/pages/HomePage/AnalysisPyWindow/AnalysisPyWindow.tsx
```diff
@@ -0,0 +1,204 @@
...
+  height,
+  onDivElement,
+}) => {
+  return (
+    <div
+      style={{ position: "absolute", width, height, overflowY: "auto" }}
+      ref={onDivElement}
+    />
+  );
+};
```
jsoules, line 156: These two components are currently identical--are they expected to diverge in the future?

gui/src/app/pages/HomePage/AnalysisPyWindow/AnalysisPyWindow.tsx
```diff
@@ -0,0 +1,204 @@
...
+  onImageOutputDiv,
+}) => {
+  return (
+    <ImageOutputWindow
+      width={width}
+      height={height}
+      onDivElement={onImageOutputDiv}
+    />
+  );
+};
```
jsoules, line 176: I'm also unclear on the function of this as an intermediary component

magland:
> Right now the UI lets you click over to analysis.py/analysis.R when you have not yet run sampling. Moving those tabs down to the area we only show after sampling has completed seems like a good idea to prevent this

Okay I have reorganized the tabs. I do think it's important to be able to view and even the analysis scripts even prior to the completion of sampling. So those tabs are present. But if you try to execute the scripts before sampling is complete, you get a (helpful?) error message.

image

jsoules (commented):
A few questions and suggested areas to tighten.

gui/src/app/Project/ProjectQueryLoading.ts
```diff
@@ -37,6 +38,7 @@ export const fromQueryParams = (searchParams: URLSearchParams) => {
     project: searchParams.get(QueryParamKeys.Project),
     stan: searchParams.get(QueryParamKeys.StanFile),
     data: searchParams.get(QueryParamKeys.DataFile),
+    'analysis.py': searchParams.get(QueryParamKeys.AnalysisPyFile),
```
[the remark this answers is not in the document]
magland: Yeah, this is an important choice (not just internal) because it's they key for the query string. I personally think analysis.py=... is the most intuitive here, even if it has a ".". Just preparing for analysis.r, data.py, data.r, and others (maybe samples.csv or initial_values.csv, etc)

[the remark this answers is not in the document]
WardBrian: I think for a query string it can be a bit long form, I suggest “python_analysis”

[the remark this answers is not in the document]
magland: If we go that route, then we're going to continue having to think of special query names, when we already have a well-defined export to project. I still don't see the downside of using those exported file names in the query string. Then we only need one convention for these things, rather than unnecessary mappings.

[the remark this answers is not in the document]
WardBrian: I personally find it confusing to see a “.” character in that part of a URL — I honestly didn’t think it would even be legal the first time I saw it

We already have mappings for all the other keys, which are not exactly their file name

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -143,6 +136,60 @@ const HomePageChild: FunctionComponent<Props> = ({ width, height }) => {
...
+        height={height}
+        compiledMainJsUrl={compiledMainJsUrl}
+      />
+      <AnalysisPyWindow width={width} height={height} />
+      <div>
+        R Analysis not yet implemented
+      </div>
+    </TabWidget>
+  );
+};
```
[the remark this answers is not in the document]
magland: The tabs have since been reorganized in response to Brian's suggestion. So I'll resolve this particular conversation, although I do agree that we should start splitting components out of some of the larger files.

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
```diff
@@ -0,0 +1,169 @@
...
+    }
+    if (!analysisPyWorker) {
+      throw new Error("analysisPyWorker is not defined");
+    }
+    if (consoleOutputDiv) {
+      consoleOutputDiv.innerHTML = "";
+    }
+    if (imageOutputDiv) {
+      imageOutputDiv.innerHTML = "";
+    }
```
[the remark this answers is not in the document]
magland: I could imagine a case (if this component is being reused somewhere else) where we might only be interested in console output (if the script is not expected to generate images) or only image output (if we don't need to show the console).

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    if (label) {
+      ret.push({
+        type: "text",
+        label,
+        color,
+      });
+    }
+    return ret;
+  }, [fileContent, editedFileContent, handleRun, status, imageOutputDiv]);
```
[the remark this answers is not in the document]
magland: I agree, but I think we should hold off until things solidify more, since we don't yet know exactly what will be in common (or different) between the 4 cases.

gui/src/app/pages/HomePage/AnalysisPyWindow/AnalysisPyWindow.tsx
magland: Yeah I think it's likely they will since the type of content is very different between them. Right now we construct the elements that go in these using non-react DOM-manipulation, but in the future we might be able to figure out how to set the content as props/state somehow.

gui/src/app/pages/HomePage/AnalysisPyWindow/AnalysisPyWindow.tsx
magland: It's just a crude way to separate layout concerns from the functional components. At this point I think the layout choices are still in flux.

gui/src/app/Project/ProjectQueryLoading.ts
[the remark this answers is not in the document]
magland: May I suggest analysis_py, analysis_r, etc? Then we're not using "." and the mapping is straightforward.

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
```diff
@@ -0,0 +1,169 @@
...
+        consoleOutputDiv?.appendChild(divElement);
+      },
+      onStderr: (x) => {
+        console.error(x);
+        const divElement = document.createElement("div");
+        divElement.style.color = "red";
+        const preElement = document.createElement("pre");
+        divElement.appendChild(preElement);
+        preElement.textContent = x;
+        consoleOutputDiv?.appendChild(divElement);
```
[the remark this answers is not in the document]
magland: This looks good to me. I'll make the change.

gui/src/app/Project/ProjectQueryLoading.ts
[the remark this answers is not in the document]
WardBrian: Fine by me - @jsoules?

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
```diff
@@ -0,0 +1,169 @@
...
+    PyodideWorkerInterface | undefined
+  >(undefined);
+
+  // worker creation
+  useEffect(() => {
+    const worker = PyodideWorkerInterface.create("analysis.py", {
+      onStdout: (x) => {
+        console.log(x);
+        const divElement = document.createElement("div");
+        divElement.style.color = "blue";
```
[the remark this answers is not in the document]
magland: Made this change.

gui/src/app/Project/ProjectQueryLoading.ts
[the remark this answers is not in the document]
jsoules: Oh yeah, sounds great here.

magland:
Thinking about the interface for analysis.py to get the draws data.

We want something that can be made backward compatible so that as we make improvements to the interface, existing SP Projects will continue to function, at least to a reasonable degree.

The only way I can think of for doing that is to provide some python imports that provide a specific interfaces, and the user can select which they want to use.

So here's an example analysis.py

```python
from sp_util import load_draws_v1

draws = load_draws_v1()
a = draws.get_dataframes()
print(a[0])

plt.figure()
plt.hist(a[0]['lp__'])
plt.show()
```

Here, by virtue of this specific import the user gets a particular interface, with the expectation that it will be maintained in a backward-compatible manner.

I prefer this import method compared with injecting a mysterious variable into the mix, because python isn't supposed to work that way. I think analysis.py should be a well-defined python unit (script) that just depends on the packages it is importing.

In the implementation from my latest commit, the sp_util module does not come from pypi... but rather it is hard-coded in pyodide... so it's internal to the stan-playground web app. The actual sampling data (draws, parameter names, etc) are communicated behind-the-scenes by writing to a file that is available to the sp_util module. The load_draws_v1 function first loads the data from that file and then prepares an object that has the get_dataframes() method. It also has a `draws.get_dataframe_longform()` method.

magland:
@WardBrian Are you suggesting the script would instead be the following, where "draws" is the special variable?

```bash

# Get a list of dataframes, one for each chain
a = draws.get_dataframes()
print(a[0])

# Plot a histogram of `lp__`
plt.figure()
plt.hist(a[0]['lp__'])
plt.show()
```

WardBrian:
Something like this, yes

magland:
> Something like this, yes

magland:
@WardBrian @jsoules 

Implemented draws object for analysis.py based on what we had discussed. Here's a gist to try

[http://localhost:3000?project=https://gist.github.com/magland/d811a59035037dd0e19e73900048a8fb](http://localhost:3000/?project=https://gist.github.com/magland/d811a59035037dd0e19e73900048a8fb)

The analysis.py for this example, which illustrates the functionality.

```python
import numpy as np

# Histograms
for pname in ['gamma', 'beta', 'phi_inv', 'lp__']:
    plt.figure()
    plt.hist(draws.get(pname))
    plt.title(pname)
    plt.show()

print('PARAMETERS')
print('==========')
for p in draws.parameter_names:
    print(p)

print('')
print('PARAMETER VALUES')
print('================')
print('gamma:', draws.get('gamma'))
print('beta:', draws.get('beta'))
print('phi_inv:', draws.get('phi_inv'))

print('')
print('OTHER')
print('=====')
print('Mean of lp__:', np.mean(draws.get('lp__')))
print('Shape of y:', draws.get('y').shape)

print('')
print('DATAFRAME')
print('=========')

df = draws.as_dataframe()
print(df)
```

Note that y is a matrix parameter, but it gets flattened to a vector. Will want to use stanio instead, as noted in the comments in the source code. @WardBrian do you want to work on this? Here's the relevant script that gets loaded into the worker and then into the pyodide environment

https://github.com/flatironinstitute/stan-playground/pull/120/files#diff-356c7427b0952fcfd57ddd11e46493a6624870c92e611c335ee52f142272654d

magland:
Assuming the structure of this looks good... In terms of merging order, it would help a lot to merge a basic version of this to minimize conflicts with other features we are working on (e.g., data generation). And then reopen a new PR with the needed tweaks to this one. What do you think?

gui/src/app/SamplerOutputView/SamplerOutputView.tsx
```diff
@@ -14,7 +14,7 @@ import TracePlotsView from "./TracePlotsView";
 type SamplerOutputViewProps = {
   width: number;
   height: number;
-  sampler: StanSampler;
+  sampler: StanSampler | undefined;
```
WardBrian, line 17: This seems like an unexpected change for this PR -- what necessitates it?

gui/src/app/pages/HomePage/AnalysisPyWindow/AnalysisPyWindow.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+const getScriptHeaderForEmptyDraws = () => {
+  return `
+raise Exception("You must run the sampler before executing the analysis script.")
+`;
+};
```
WardBrian, line 190: Is this necessary if we're going to block the UI?

gui/public/sp_load_draws.py
```diff
@@ -0,0 +1,67 @@
...
+from typing import List
+import numpy as np
+
+
+class DrawsObject:
+    def __init__(self, sp_data):
+        if "sampling" not in sp_data:
+            raise ValueError("sampling key not found in sp_data")
+        sampling: dict = sp_data["sampling"]
+        self._draws: List[List[float]] = sampling["draws"]
```
WardBrian, line 12: I think it would make sense to have the internal representation use numpy starting in the constructor, rather than doing it on demand later. I think all the ways we want to expose it will prefer the data already be an array

gui/src/app/pyodide/pyodideWorker/pyodideWorker.ts
```diff
@@ -0,0 +1,246 @@
...
+          await pyodide.loadPackage("micropip");
+          const microPip = pyodide.pyimport("micropip");
+          await microPip.install("arviz<0.18");
+        } finally {
+          setStatus("running");
+        }
+      }
+      pyodide.FS.writeFile("_sp_data.json", JSON.stringify(spData), {
+        encoding: "utf-8",
+      });
```
WardBrian, line 136: Writing this to a file seems a bit odd to me compared to initializing it as a global. Is there an advantage to this approach?

gui/src/app/pyodide/pyodideWorker/pyodideWorker.ts
```diff
@@ -0,0 +1,246 @@
...
+    });
+    if (pyodideWorkerMode === "data.py") {
+      const micropip = pyodide.pyimport("micropip");
+      await micropip.install("stanio");
+    }
+    if (pyodideWorkerMode === "analysis.py") {
+      const sp_load_draws_py = await fetchScript('/sp_load_draws.py')
+      pyodide.FS.writeFile("sp_load_draws.py", sp_load_draws_py, {
+        encoding: "utf-8",
+      });
```
WardBrian, line 41: Again with written to file vs included in preamble

gui/src/app/pyodide/pyodideWorker/pyodideWorker.ts
```diff
@@ -0,0 +1,246 @@
...
+  if (pyodide === null) {
+    if (!pyodideWorkerMode) {
+      throw Error("pyodideWorkerMode is not defined");
+    }
+    const packages =
+      pyodideWorkerMode === "data.py"
+        ? ["numpy", "micropip"]
+        : pyodideWorkerMode === "analysis.py"
+          ? ["numpy", "matplotlib", "pandas"]
+          : [];
```
WardBrian, line 22: I think we can probably branch on the mode less often than we are. Might need to wait until implementing the data version to be sure

WardBrian (changes requested):
A few things stood out to me in the current version -

gui/public/sp_load_draws.py
```diff
@@ -0,0 +1,64 @@
...
+            if len(pp) == 0:
+                raise ValueError(f"Parameter {pname} not found")
+            # Here we assume that the parameters are in the correct order.
+            # Matrices will be flattened.
+            # TODO: Use stanio for this instead
+            inds = [self._all_parameter_names.index(p) for p in pp]
+            return np.array([[draw[ind] for ind in inds] for draw in self._draws])
+
+    @property
+    def parameter_names(self):
```
jsoules, line 53: Where is this intended to be used? I couldn't find it elsewhere in the code.

gui/public/sp_load_draws.py
```diff
@@ -0,0 +1,64 @@
+# Used in pyodideWorker for running analysis.py
+
+from typing import List
+import numpy as np
+
+
+class DrawsObject:
+    def __init__(self, sp_data):
```
jsoules, line 8: I'm sorry, I don't remember what `sp_data` is supposed to be or what assumptions we can make about it. Also, is there any type hinting here that would be appropriate?

Also it seems like we're only interested in the `sampling` member, should we narrow the parameter to just that?

jsoules (commented):
Some initial thoughts--may have more as I dig more.

gui/src/app/SamplerOutputView/SamplerOutputView.tsx
magland: It's because we now need the output view to exist even when the stan program has not been compiled. It's one of the three tabs.

gui/src/app/pages/HomePage/AnalysisPyWindow/AnalysisPyWindow.tsx
magland: You're right, I haven't blocked the run button in the UI yet. I'll do that now.

gui/src/app/pyodide/pyodideWorker/pyodideWorker.ts
magland: I think it's cleaner as a file because then any of the scripts can access it without needing to pass it around as a variable. And I think it's good to keep the main script as simple as possible and put more complex logic in the sp_load_draws.py

gui/src/app/pyodide/pyodideWorker/pyodideWorker.ts
magland: I'm thinking of possibly having other files being imported in, and it's better to have them be stand-alone and imported in the main script. Again, trying to keep the main script as simple as possible.

magland: Some files may need to import others.

gui/public/sp_load_draws.py
```diff
@@ -0,0 +1,64 @@
...
+        import pandas as pd
+
+        data = []
+        for chain_id in range(1, self._num_chains + 1):
+            chain_draws = []
+            for i in range(len(self._draws)):
+                if self._chain_ids[i] == chain_id:
+                    chain_draws.append(self._draws[i])
+            for draw_index, draw in enumerate(chain_draws):
+                data.append([chain_id, draw_index + 1] + draw)
```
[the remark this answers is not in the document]
magland: Yes this is intentional, because then we can fill in the second column (draw index) appropriately.

gui/public/sp_load_draws.py
magland: The user might want to get a list of the parameter names.

gui/public/sp_load_draws.py
magland: sp_data in Python corresponds to the spData on the javascript side. It's the way that the sampling draws are passed from js -> pyodide. Typing a dict is a lot easier in typescript than in Python. I didn't want to bring in pydantic or something heavy.

magland: I thought we might need to provide more information than just sampling in the future.

gui/public/sp_load_draws.py
magland: Right now the other parts of this class assume it's a Python list. Should we rewrite it to use numpy arrays? I'm inclined to leave it as is, but if you think there's an important reason.

gui/src/app/pyodide/pyodideWorker/pyodideWorker.ts
magland: And the more preamble we have, the more names are contaminating the namespace...

gui/src/app/pages/HomePage/AnalysisPyWindow/AnalysisPyWindow.tsx
magland: The UI now does not allow running when there are no samples, and there's a helpful message on the editor.

gui/public/sp_load_draws.py
magland: I have now added some additional typing on the typescript side

```typescript
export type GlobalDataForAnalysisPy = {
  sampling?: {
    draws: number[][];
    paramNames: string[];
    numChains: number;
    chainIds: number[];
  };
};
```

gui/src/app/pyodide/pyodideWorker/pyodideWorker.ts
magland: @WardBrian I thought about it some more and I think you're right. Now it's being passed as global variable. (but I still think sp_load_draws.py should be in its own file)

gui/public/sp_load_draws.py
jsoules: I can't speak for Brian here, but one reason that occurs to me would be redundant memory use: if we store as both a list and an NDArray, we're storing the data twice, which could get burdensome for large draw counts.

Within the class, it looks like most of what we would be doing with the list would work for an `ndarray` with some fairly small modifications. Are there other places it's used that would be of concern?

magland: Fair point about the memory. If one of you wants to rewrite it using numpy internally it's fine with me, but I prefer it as is so I wouldn't want to go to the effort myself.

gui/public/sp_load_draws.py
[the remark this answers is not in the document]
jsoules: Okay. But it seems like one could do it in one pass?

Maybe I've been thinking wrongly about `self._draws`. This is actually a 2d array, right? So `len(self._draws)` is actually the number of chains, and the outer loop is just iterating over chains.

So in
```python
for i in range(len(self._draws)):
    if self._chain_ids[i] == chain_id:
        chain_draws.append(self._draws[i])
```
this is going to run n = num_chains times; each iteration will do `chain_draws.append` exactly once, to load an item that is "all the draws for the chain". Does that sound right?

Wouldn't
```python
data = []
for i in range(1, self._num_chains + 1):
    for draw_index, draw in enumerate(self._draws[i]):
        data.append([self._chain_ids[i], draw_index + 1] + draw)
```
accomplish the same thing as this section?

magland:
I have merged in recent changes from main and resolved some conflicts. In particular the non-relative imports.

WardBrian:
@jsoules @magland this PR is now updated with respect to main. I've done a basic once-over to make sure everything still works as expected, but it can now be given a more thorough review!

magland:
@WardBrian Nice! I have tested it and everything seems to be working.

As I mentioned I think we should do a separate PR for analysis.r and we can borrow code from #94 . LMK if you want me to work on that -- I can draft an initial version, but I will need help figuring out how to make the R draws object analogous to the python one.

gui/src/app/pages/HomePage/DataGenerationWindow/DataPyFileEditor.tsx
```diff
@@ -0,0 +1,103 @@
...
+  fileName,
+  fileContent,
+  onSaveContent,
+  editedFileContent,
+  setEditedFileContent,
+  setData,
+  readOnly,
+  outputDiv,
+}) => {
+  const [status, setStatus] = useState<PyodideWorkerStatus>("idle");
```
jsoules, line 35: Funny that I never thought about this before, but is it weird that the worker status is something tracked by a main-thread UI component rather than the component querying the worker for its status?

I'm sure this actually makes sense (esp. b/c of the worker thread going nonresponsive while it's actually running a job) but it just stuck out to me.

jsoules (approved):
I think this is overall in good shape. I have a few minor notes but would defer anything more substantial to a later refactor.

We should be able to get this merged in short order I think.

jsoules:
Oh, one thing that would be nice that I didn't mention because it isn't strictly germane to this PR, but it would be good for us to document somewhere the expectations around the global data/memory sharing model that's used to pass data around among the workers/non-TS, non-Stan interpreters and our own app. Like, what's the data structure look like, what's the strategy.

gui/src/app/pages/HomePage/DataGenerationWindow/DataPyFileEditor.tsx
magland: I know what you mean, but I think it's important for the status state to be synchronously available in the main thread.

gui/src/app/SamplerOutputView/TracePlotsView.tsx
```diff
@@ -86,6 +86,7 @@ const SequencePlotChild: FunctionComponent<SequencePlotProps> = ({
   }, [draws, columnIndex, drawChainIds]);
   return (
     <div className="SequencePlotChild">
+      {/* TODO this is raising deprecation warnings */}
```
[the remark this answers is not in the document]
WardBrian: Replaced with the MUI Accordion component

gui/src/app/pages/HomePage/DataGenerationWindow/DataGenerationWindow.tsx
```diff
@@ -0,0 +1,98 @@
+import { FunctionComponent, useCallback, useContext, useRef } from "react";
+import DataPyFileEditor from "./DataPyFileEditor";
+import { ProjectContext } from "../../../Project/ProjectContextProvider";
+import { ProjectKnownFiles } from "../../../Project/ProjectDataModel";
+import { ConsoleOutputWindow } from "../AnalysisPyWindow/AnalysisPyWindow";
+import DataRFileEditor from "./DataRFileEditor";
+import { writeConsoleOutToDiv } from "app/pyodide/AnalysisPyFileEditor";
```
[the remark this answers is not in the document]
magland: Fixed.

gui/src/app/pages/HomePage/DataGenerationWindow/DataRFileEditor.tsx
```diff
@@ -0,0 +1,156 @@
+import {
+  FunctionComponent,
+  RefObject,
+  useCallback,
+  useMemo,
+  useState,
+} from "react";
+import { WebR } from "webr";
+import TextEditor, { ToolbarItem } from "../../../FileEditor/TextEditor";
+import getDataGenerationToolbarItems from "./getDataGenerationToolbarItems";
```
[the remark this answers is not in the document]
magland: Fixed

gui/src/app/pages/HomePage/DataGenerationWindow/getDataGenerationToolbarItems.tsx
```diff
@@ -0,0 +1,58 @@
+import { ToolbarItem } from "@SpComponents/TextEditor";
+import { Help, PlayArrow } from "@mui/icons-material";
+import { PyodideWorkerStatus } from "app/pyodide/pyodideWorker/pyodideWorkerTypes";
```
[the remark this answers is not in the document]
magland: Fixed

gui/src/app/FileEditor/TextEditor.tsx
```diff
@@ -15,10 +14,72 @@ import {
...
+          },
+        ],
+        [/~|:|\+=|=|<|>/, "stanoperator"],
+        ...stanLang.tokenizer.root,
+      ];
+      monacoInstance.languages.register({ id: "stan" });
+      monacoInstance.languages.setMonarchTokensProvider("stan", stanLang);
+    });
+  }
+});
```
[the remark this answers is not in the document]
WardBrian: This was already in the file further down, but moving it out of the component fixed some rerender flickers. I've moved it into the StanLang file

gui/src/app/pages/HomePage/SamplingWindow/SamplingWindow.tsx
```diff
@@ -0,0 +1,88 @@
...
+  modelHasUnsavedDataFileChanges,
+} from "../../../Project/ProjectDataModel";
+import RunPanel from "../../../RunPanel/RunPanel";
+import SamplerOutputView from "../../../SamplerOutputView/SamplerOutputView";
+import SamplingOptsPanel from "../../../SamplingOptsPanel/SamplingOptsPanel";
+import useStanSampler, { StanRun } from "../../../StanSampler/useStanSampler";
+import AnalysisPyWindow from "../AnalysisPyWindow/AnalysisPyWindow";
+import Box from "@mui/material/Box";
+import Grid from "@mui/material/Grid";
+import Divider from "@mui/material/Divider";
```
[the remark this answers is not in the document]
WardBrian: I'm still pretty used to Java or C/C++ which make this style look absolutely frugal by comparison...

gui/src/app/pyodide/pyodideWorker/pyodideWorkerTypes.ts
```diff
@@ -0,0 +1,76 @@
...
+  code: string;
+  spData: Record<string, any> | undefined;
+  spRunSettings: PyodideRunSettings;
+};
+
+export const isMessageToPyodideWorker = (
+  x: any,
+): x is MessageToPyodideWorker => {
+  if (!x) return false;
+  if (typeof x !== "object") return false;
```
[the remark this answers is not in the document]
magland: Done.

magland:
@WardBrian it looks like we had some simultaneous commits, doing some of the same things. Hopefully everything shakes out properly.

WardBrian:
Yep, I resolved any conflicts

jsoules:
Ok, looks to me like we're good to merge this--everybody good with that?

WardBrian (approved):

PR #121, opened 2024-07-03
Log stanc version on load

WardBrian:


jsoules (approved):
Changes appear straightforward enough. Tried it out in the browser and it does in fact log the stanc version.

Ship it

PR #124, opened 2024-07-09
Minor changes to type guards, and adding tests.

jsoules:
This implements the changes suggested in #93 (too late for inclusion in that PR).

Specifically, I've added a minor utility function to reduce the repeated basic checks in the type guards, as well as implementing more detailed deserialization-time validity checking for the sampling options.

I've also added some (pedantic, not very flexible) tests for the type guards implemented in that PR, as well as a sketch of some for the query parameter loading file (not complete coverage, since I left out the long function which was--hopefully--out of scope for testing with the rest of the changes I was making).

I've also reorganized a little bit to move some utility functions into common areas, and to (somewhat) reduce the fragmentation of types & type guards across the code (by moving the sampling options definition into the `ProjectDataModel` file--though perhaps there's a better place for these?)

I do not know why my `Prettier` format-on-save insists on resetting to 4 spaces for imports...


WardBrian:
> I do not know why my Prettier format-on-save insists on resetting to 4 spaces for imports...

Are you sure vscode is using Prettier? If you press F1 and search for format, it might complain there are multiple, and then let you pick Prettier

gui/test/app/Project/ProjectQueryLoading.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    test("Returns false if query object has no non-null elements", () => {
+      const emptyQuery: any = {};
+      expect(queryStringHasParameters(emptyQuery)).toEqual(false);
+    });
+  });
+});
+
+// TODO: Testing fetchRemoteProject
```
WardBrian, line 81: Want to do this now, or?

gui/src/app/Project/ProjectDataModel.ts
```diff
@@ -1,21 +1,69 @@
...
+  const naturalFields = [x.num_chains, x.num_samples, x.num_warmup];
+  const positiveFields = [x.num_chains, x.num_samples];
+  const nonnegativeFields = [x.num_warmup, x.init_radius];
+  if (naturalFields.some((f) => !Number.isInteger(f))) return false;
+  if (positiveFields.some((f) => f <= 0)) return false;
+  if (nonnegativeFields.some((f) => f < 0)) return false;
+  return true;
+};
+
+export const parseSamplingOpts = (x: string | undefined): SamplingOpts => {
```
WardBrian, line 38: Should we support partial fields? My thought is if we ever expose additional arguments, old links could break if we're insisting on a full parse.

I know we don't currently, so it's fine to not do this now, just speculating

gui/test/app/Project/ProjectQueryLoading.test.ts
WardBrian: Or is this the long function you mentioned intentionally skipping

WardBrian (approved):
Looks good! A few comments

gui/test/app/Project/ProjectQueryLoading.test.ts
jsoules: That's correct. It ought to be tested, but I really only touched this file to use the new parsing function, so I realized I was getting too far into the weeds; yet didn't want to delete the work I'd done already.

gui/src/app/Project/ProjectDataModel.ts
jsoules: Yeah, it's a good question. So far we appear to be insisting on it, but fortunately for now the only way to get one of these (other than to craft it manually, which, why) would be to save one, and as it currently stands that will give a full object.

...but going forward, yeah, we should probably drop in defaults for non-provided values.

PR #129, opened 2024-07-12
Update to use path aliases and resolver.

jsoules:
This does away with relative imports, which should help future-proof our project through reorganization, and also shortens a lot of the imports quite a bit.

Needed to include a path resolver (for both the main vite config and a worker-specific one) in order to get the build system to recognize the import aliases.

Since there's a lot of files with small changes, there will be a bit of retrofitting required for the imports on our other outstanding PRs.

We may wish to use different aliases or a different alias structure--I prefixed everything so they'd all line up in the same area.

Also note, working out the aliases suggested that there's a lot of files that could do for some file/directory structure reorganization for where the files actually live as well. We can look at that later.


gui/package.json
```diff
@@ -38,7 +38,8 @@
     "react-plotly.js": "^2.6.0",
     "react-router-dom": "^6.17.0",
     "react-visibility-sensor": "^5.1.1",
-    "tinystan": "^0.0.2"
+    "tinystan": "^0.0.2",
+    "vite-tsconfig-paths": "^4.3.2"
```
WardBrian, line 42: Can this not be a dev- only dependency?

gui/src/app/FileEditor/TextEditor.tsx
```diff
@@ -1,6 +1,9 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import { Hyperlink, SmallIconButton } from "@fi-sci/misc";
 import { Editor } from "@monaco-editor/react";
 import { Refresh, Save } from "@mui/icons-material";
+import { highlightJsData } from "@SpComponents/stanLang";
```
WardBrian, line 5: This import coming from the Components alias strikes me as odd

gui/tsconfig.json
```diff
@@ -19,7 +19,33 @@
...
+      "@SpComponents/*": [
+        /* This implies a certain reorganization/recollection process */
+        "app/components/*",
+        "app/FileEditor/*",
+        "app/RunPanel/*",
+        "app/SamplerOutputView/*",
+        "app/SamplingOptsPanel/*",
+        "app/TabWidget/*"
+      ],
+      "@SpCore/*": ["app/Project/*", "app/gists/*"],
```
WardBrian, line 36: Is it possible to keep sub-paths in this setup?

For example, I think I would ideally have everything under Project/ directly under SpCore, but keep the /gist/ in the path for those files, like SpCore/gist/...

WardBrian (commented):
This was overdue, thanks! 

A few comments

gui/package.json
jsoules: Probably--I probably just installed it wrong. Let me change it and see if it still builds.

gui/src/app/FileEditor/TextEditor.tsx
jsoules: I can see that--just a byproduct of where the file is actually located within the project. It might make more sense (since this is defining a bunch of language keywords) to put it in `Stanc`?

gui/tsconfig.json
jsoules: This is doable (and now done). Happy to look at other such things as well.

gui/src/app/FileEditor/TextEditor.tsx
WardBrian: Maybe -- I want to re-vamp the highlighting eventually. Then it might make sense to have a 'stan language' folder with 'compiler' as a sub-folder or something. Nothing for now, I think.

WardBrian (approved):

PR #128, opened 2024-07-12
Add tests for project data reducer.

jsoules:

PR #130, opened 2024-07-15
Replace 'stanies' with links

WardBrian:
We discussed this a few places while implementing gist loading, and before we invested in testing it in #128 it seemed like worth doing.

I also updated the default sampling options of the disease transmission to give more reasonable runtimes.

~Note: a side effect of doing it like this is the entire UI reloads. It's nice that now the links can be right clicked and copied, but we could also think of different ways which wouldn't give true links but would still store the examples outside a JS source file.~


WardBrian:
Gists for organizations is a request dating back [until at least 2010](https://github.com/orgs/community/discussions/7923) with no progress.

We could also store them in either the public folder and link to them that way or in the repo and use github's "raw" links, but gist is currently the most compact way we have.

jsoules:
Not strictly germane to this PR, but I'm noticing that we have doubled entries in the history list when loading these, and only the chronologically-first of them actually loads the relevant gist.

WardBrian:
That's more-or-less intentional - the more recent of the two is what gets set after the search parameters are cleared (if you hover over you should see one has the param, one doesnt), and represents the possibly-dirty window that can't be revisted from a history menu. Unless you're seeing duplicates beyond that?

jsoules:
I think I approve this, assuming there are no more changes to it.

WardBrian:
@jsoules I just pushed a small change to try to make the histories clearer

jsoules:
@WardBrian Certainly makes the history entries clearer. It can be a little confusing that the "Editing XXX" link does not actually load XXX, but I think this is an improvement and probably the best we're going to do barring a larger rework, so I wouldn't put anything more into this right now.

PR #131, opened 2024-07-15
Set example links to replace current history item

WardBrian:
A very minor improvement which occurred to me mere seconds after merging #130 

This makes it so clicking the examples replaces the current history item, which should always be the empty URL with no parameters (and therefore a relatively uninteresting history item). This makes it so clicking the various examples multiple times leads to a much more useful history/back button listing.


jsoules:
Would seem to work a treat. Ship it.

Issue #112, opened 2024-06-29
Simplify div absolute positioning and styling


magland:
Since we are at a good point for sweeping through the codebase and making changes, I wanted to propose the following for simplifying all the absolute positioning and styling of divs throughout. This doesn't preclude doing a further overhaul down the road, but this should tidy things up a bit.

Put a "Div" component, in the components/ folder defined like this

```typescript
import { CSSProperties, FunctionComponent, PropsWithChildren } from "react"

type DivProps = {
    left?: number | undefined
    top?: number | undefined
    width?: number | undefined
    height?: number | undefined
    position?: 'absolute' | 'relative'
    xScroll?: boolean
    yScroll?: boolean
    backgroundColor?: string
    style?: CSSProperties
    onKeyDown?: (event: React.KeyboardEvent<HTMLDivElement>) => void
    className?: string
}

const Div: FunctionComponent<PropsWithChildren<DivProps>> = ({
    children,
    left = undefined,
    top = undefined,
    width = undefined,
    height = undefined,
    position = 'absolute',
    xScroll = false,
    yScroll = false,
    backgroundColor = undefined,
    style = {},
    onKeyDown = undefined,
    className = undefined
}) => {
    const divStyle: CSSProperties = {
        left,
        top,
        width,
        height,
        position,
        overflowX: xScroll ? 'scroll' : 'hidden',
        overflowY: yScroll ? 'scroll' : 'hidden',
        backgroundColor,
        ...style
    }
    return (
        <div className={className} style={divStyle} onKeyDown={onKeyDown}>
            {children}
        </div>
    )
}

export default Div
```

Then for example

```typescript
<div style={{position: 'absolute', width, height, overflow: 'auto'}}>
```

Would become

```typescript
<Div width={width} height={height} yScroll={true}>
```

and

```typescript
<div className="left-panel" style={{ position: 'absolute', left: 0, top: topBarHeight + 2, width: leftPanelWidth, height: height - topBarHeight - 2, overflow: 'auto' }}>
```

would become

```typescript
<Div className="left-panel" width={leftPanelWidth} height={height - topBarHeight - 2} top={topBarHeight + 2} yScroll={true}>
```

magland:
> If I’m reading this correctly it sounds like an alternative to #23 and #24, correct?
> 
> I think I would prefer using css classes that live outside the code files for as much as possible. Obviously if we need a computed value like a width we’re feeding in from the outside that may be a different story

[...]

PR #133, opened 2024-07-16
Replace inline styling with CSS classes where possible.

jsoules:
An attempt to use semantically-meaningfully-named CSS classes over inline styling. Most styling is in `localStyles.css` at the top level, though I'm not too attached to that; it might make more sense to break them down by the area of the app they impact, but our situation isn't sufficiently complex to justify an elaborate CSS solution right now in my mind.

Some class names may be subject to revision if I've misconstrued the function they serve in the actual layout. There are also a few cases where I'd like to combine some of the stylng in separate css files into `localStyles.css`, but I wanted to discuss a bit first whether that's the right thing to do.

One area to remark is that we have style sheets for several different types of tables with no clear distinction about what's used where. I removed the `draws-table-2.css` stylesheet since no existing component was actually using any of it, and I renamed `table1.css` to something a bit more semantic that I think identifies its function, but there's more that could be done here.

Many instances of inline styling still exist; these are mainly either manually computed positioning for elements, or including a "color" style as a parameter for one of the toolbars. I think we would benefit from moving away from absolute positioning & manually computed dimensions, but that's outside the scope of the current PR.


WardBrian (approved):
Looks good!

magland:
@WardBrian @jsoules 
I have concerns about this. I recognize that it's best to put styling in external .css files (and we should do that), but it bothers me that we're maintaining the styles globally. I understand that global styling makes sense when we're talking about colors or padding that apply to multiple components. But what you are doing here is is making global entries for specific components, IMO breaking the modularity of the component system.

Also see this:
https://github.com/flatironinstitute/stan-playground/issues/112

I think there is a difference between layout concerns using css and styling concerns (colors, etc). And those should be treated separately.

Since this is already merged I think we can go forward, but I may propose reverting a lot of this to address the above concerns.

PR #134, opened 2024-07-17
Simplify useStanSampler and related hooks

WardBrian:
This collapses a lot of the state management around the Stan model worker and its output into a reducer which is managed internally.

useStanSampler has two concerns:

1. Creating the worker thread, allowing us to call it, and managing cancellation/cleanup
2. Providing reactive access to the outputs of sampling

After this PR, these now exactly correspond to the two returned objects from the hook, `sampler` and `lastRun`. The later is a simple POD struct which supplants the existing useSamplerStatus, useProgress, and useSamplerOutputs hooks.


magland:
At first glance it looks good to me. How do you feel about including the state as part of the "ephemera" of the project?

WardBrian:
I'd be slightly worried about sending that many updates to the global project data (through e.g. the progress updates). But, assuming React was able to avoid that leading to unnecessary work, it still seems cleaner to keep it separate since it is not being persisted, shared, etc

jsoules (approved):
I think these changes make the logic a lot more straightforward.

PR #141, opened 2024-07-23
Don't save empty files to zip

WardBrian:
Particularly now that we have data and analysis scripts we can expect some of them to be empty. The Gist feature already does this filtering


magland:
Looks good to me.

Issue #142, opened 2024-07-23
Make pyodide worker a global


WardBrian:
This would fix the current frustration that closing then reopening a tab with pyodide inside means it needs to re-download/re-initialize

magland:
Perhaps this should be global anyway but... more fundamentally the analysis.py react component should not unmount when a new tab is selected. This is important for other reasons, such as maintaining the splitter states and the scroll position in the editor. This needs to be changed in TabWidget. I'm not certain, but I believe a previous version of TabWidget had the property of preserving the child components.

WardBrian:
You can prevent the tab widget from unmounting its children by using just `{children}` instead of `{value === index && <>{children}</>}` on line 59. I believe this has some potentially unintended effects like causing plotly to load in the background even when the tab isn't selected yet

magland:
It will take some more coding that might look hacky, but I think we should have tab widget load tabs only when they are first opened, and then remember which had been opened. Requires internal state keeping track of which tabs had been opened.

magland:
I just thought of a reason not to have a single global pyodide worker. We want to be able to kill runs, and I think that involves terminating the worker. So I think the worker should be specific to that editor window.

PR #144, opened 2024-07-24
Track viewed tabs in TabWidget component

magland:
This was discussed in #142

With this PR: In tab widget, tab content is not removed from DOM when switching between tabs, but also the tab content is not included in the DOM until the tab was viewed at least once. This involves keeping a state `tabsThatHaveBeenViewed`.

To verify this is working:

* Run sampling and click on ANALYSIS (PY) tab
* Resize the editor
* Switch to the OUPUT tab and then back to the ANALYSIS tab
* Note that the resize has persisted

Then for a second test:
* Run sampling and wait for a few seconds
* In the OUTPUT tab, click HISTOGRAMS tab
* Note that plotly is being loaded, demonstrating that it hadn't been loaded previously.

### DIFF
--- a/gui/src/app/components/TabWidget.tsx
+++ b/gui/src/app/components/TabWidget.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { FunctionComponent, useState } from "react";
+import { FunctionComponent, useEffect, useState } from "react";
 import Tabs from "@mui/material/Tabs";
 import Tab from "@mui/material/Tab";
 import Box from "@mui/material/Box";
@@ -15,6 +15,17 @@ const TabWidget: FunctionComponent<TabWidgetProps> = ({ labels, children }) => {
   }
 
   const [index, setIndex] = useState(0);
+  const [tabsThatHaveBeenViewed, setTabsThatHaveBeenViewed] = useState<
+    number[]
+  >([]);
+  useEffect(() => {
+    setTabsThatHaveBeenViewed((prev) => {
+      if (!prev.includes(index)) {
+        return [...prev, index];
+      }
+      return prev;
+    });
+  }, [index]);
 
   const handleChange = (_: React.SyntheticEvent, newValue: number) => {
     setIndex(newValue);
@@ -29,7 +40,12 @@ const TabWidget: FunctionComponent<TabWidgetProps> = ({ labels, children }) => {
       </Tabs>
       <Box flex="1" overflow="scroll">
         {children.map((child, i) => (
-          <CustomTabPanel key={i} value={index} index={i}>
+          <CustomTabPanel
+            key={i}
+            value={index}
+            index={i}
+            tabsThatHaveBeenViewed={tabsThatHaveBeenViewed}
+          >
             {child}
           </CustomTabPanel>
         ))}
@@ -41,11 +57,12 @@ const TabWidget: FunctionComponent<TabWidgetProps> = ({ labels, children }) => {
 interface TabPanelProps {
   children?: React.ReactNode;
   index: number;
+  tabsThatHaveBeenViewed: number[];
   value: number;
 }
 
 const CustomTabPanel = (props: TabPanelProps) => {
-  const { children, value, index, ...other } = props;
+  const { children, value, index, tabsThatHaveBeenViewed, ...other } = props;
 
   return (
     <div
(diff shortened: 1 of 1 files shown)


gui/src/app/components/TabWidget.tsx
```diff
@@ -15,6 +15,17 @@ const TabWidget: FunctionComponent<TabWidgetProps> = ({ labels, children }) => {
   }
 
   const [index, setIndex] = useState(0);
+  const [tabsThatHaveBeenViewed, setTabsThatHaveBeenViewed] = useState<
+    number[]
```
WardBrian, line 19: Thoughts on using a boolean[] of the same length as labels, rather than a number[]?

gui/src/app/components/TabWidget.tsx
```diff
@@ -15,6 +15,17 @@ const TabWidget: FunctionComponent<TabWidgetProps> = ({ labels, children }) => {
...
+  >([]);
+  useEffect(() => {
```
WardBrian, line 21: This logic could be added to handleChange below rather than in a useEffect

gui/src/app/components/TabWidget.tsx
```diff
@@ -29,7 +40,12 @@ const TabWidget: FunctionComponent<TabWidgetProps> = ({ labels, children }) => {
       </Tabs>
       <Box flex="1" overflow="scroll">
         {children.map((child, i) => (
-          <CustomTabPanel key={i} value={index} index={i}>
+          <CustomTabPanel
+            key={i}
+            value={index}
+            index={i}
+            tabsThatHaveBeenViewed={tabsThatHaveBeenViewed}
```
WardBrian, line 47: (Assuming the above change to boolean[] is done) I think it would be clearer if this is just a `mounted : Boolean`, rather than the child knowing about this entire array

WardBrian, line 47: (Assuming the above change to boolean[] is done) I think it would be clearer if this is just a `mounted : Boolean`, rather than the child knowing about this entire array

gui/src/app/components/TabWidget.tsx
```diff
@@ -56,7 +73,9 @@ const CustomTabPanel = (props: TabPanelProps) => {
       style={{ height: "100%", width: "100%" }}
       {...other}
     >
-      {value === index && <>{children}</>}
+      {(value === index || tabsThatHaveBeenViewed.includes(index)) && (
```
WardBrian, line 76: I don’t believe there is any circumstance where the first condition is true and the second is false, so this could be simplified to just `mounted &&` with the above change

WardBrian (commented):
A few thoughts on the implementation of this. I agree it’s the right thing to do in general

magland:
@WardBrian I made all your suggested changes/simplifications, except for your changing the state type to boolean[]. I think it's cleaner as number[] because

* More natural to initialize as []
* May want to accommodate dynamic changes to number of tabs in future (at that point we'd want to use tab IDs rather than indexes)
* State updates should create immutable objects. It's more natural to update a list compared with updating an entry in a boolean[] in an immutable way.

jsoules:
> @WardBrian I made all your suggested changes/simplifications, except for your changing the state type to boolean[]. I think it's cleaner as number[] because
> 
>     * More natural to initialize as []
> 
>     * May want to accommodate dynamic changes to number of tabs in future (at that point we'd want to use tab IDs rather than indexes)
> 
>     * State updates should create immutable objects. It's more natural to update a list compared with updating an entry in a boolean[] in an immutable way.

An additional possibility would be to just use a `string[]` whose values are the labels. If we are planning for the situation where tabs might be dynamic, we might also want to support reordering, which would be difficult with an index-based or order-based numeric ID.

But I don't really feel too strongly about this--I think it'll be easy enough to adapt the solution if we ever decide to support those features.

WardBrian:
Yeah, I'm happy with the current solution and the options we have for crossing-bridges-should-we-come-to-them

gui/src/app/components/TabWidget.tsx
```diff
@@ -15,8 +15,17 @@ const TabWidget: FunctionComponent<TabWidgetProps> = ({ labels, children }) => {
   }
 
   const [index, setIndex] = useState(0);
+  const [tabsThatHaveBeenViewed, setTabsThatHaveBeenViewed] = useState<
+    number[]
+  >([]);
```
WardBrian, line 20: the initial value needs to be `[index]` to get rendering correct on the first go-round

WardBrian (changes requested):

PR #145, opened 2024-07-24
Add vercel analytics

magland:
See https://vercel.com/docs/analytics

Importantly, it only stores anonymized data and [does not use cookies](https://vercel.com/docs/analytics#how-visitors-are-determined)

This will help us see the number of unique visitors per day, their browser, their OS, their Country. But note that these pieces of data are not stored jointly... so we just get aggregated plots.

### DIFF
--- a/gui/package.json
+++ b/gui/package.json
@@ -29,6 +29,7 @@
     "@mui/icons-material": "^5.15.17",
     "@mui/material": "^5.15.17",
     "@octokit/rest": "^21.0.0",
+    "@vercel/analytics": "^1.3.1",
     "jszip": "^3.10.1",
     "monaco-editor": "^0.48.0",
     "plotly.js-cartesian-dist": "^2.33.0",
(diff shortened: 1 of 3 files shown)


WardBrian:
I don't have a particularly strong feeling either way with this, but I just want to ask (maybe a stupid question), does this cause any problems if we have users in the EU or covered by other data protection laws?

magland:
According to this, we don't need to worry about it for Vercel
https://posthog.com/blog/best-gdpr-compliant-analytics-tools

magland:
Do you have an opinion about this @jsoules ?

jsoules:
Not a strong one. I've used Google Analytics in the past, but that does involve cookies and opt-in for GDPR areas. I don't particularly relish being more tightly bound to the Vercel ecosystem, but I suppose it's easy enough to back out later if we decide that's what's best.

So yeah, I think this is a good idea.

PR #147, opened 2024-07-24
Parse partial sampling opts, fill remaining keys with defaults

### LINKED ISSUE #125: Allow partial parse for SamplingOpts
As mentioned in comments for PR #124 

Currently we insist on a full parse for a `SamplingOpts` object to be considered successfully deserialized. Going forward it will likely be more resilient and flexible to populate any missing values with defaults instead.

WardBrian:
Closes #125

### DIFF
--- a/gui/src/app/Project/ProjectDataModel.ts
+++ b/gui/src/app/Project/ProjectDataModel.ts
@@ -33,13 +33,16 @@ const validateSamplingOpts = (x: SamplingOpts): boolean => {
   if (naturalFields.some((f) => !Number.isInteger(f))) return false;
   if (positiveFields.some((f) => f <= 0)) return false;
   if (nonnegativeFields.some((f) => f < 0)) return false;
+
+  if (Object.keys(x).length !== 5) return false;
   return true;
 };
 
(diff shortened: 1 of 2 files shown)


gui/src/app/Project/ProjectDataModel.ts
```diff
@@ -33,13 +33,16 @@ const validateSamplingOpts = (x: SamplingOpts): boolean => {
   if (naturalFields.some((f) => !Number.isInteger(f))) return false;
   if (positiveFields.some((f) => f <= 0)) return false;
   if (nonnegativeFields.some((f) => f < 0)) return false;
+
+  if (Object.keys(x).length !== 5) return false;
```
jsoules, line 37: Not sure I love hard-coding the 5; it isn't likely to be an issue in a serious way but I can see us tripping over this every time we change the number of potential sampling opts (which admittedly isn't going to be often).

jsoules (approved):
Code changes are modest and look good to me. Tested by saving a project, deleting some keys, and reloading, which confirmed the deleted keys were replaced with defaults.

PR #148, opened 2024-07-24
Improve data.R integration

WardBrian:
Three small things:

1. Don't display "Loading pyodide" when we are in fact loading webr
2. Check to make sure `data` is a list before serializing (`data` is also the name of a built-in R function, which we were serializing before if you just run a blank file)
3. Call [webr::shim_install()](https://docs.r-wasm.org/webr/latest/packages.html#shimming-install.packages-for-webr)

The following script can now re-create the data in the disease transmission example:

```R
install.packages("outbreaks")

library(outbreaks)

N <- 763 # given in documentation

cases <- influenza_england_1978_school$in_bed
n_days <- length(cases)
ts <- 1:n_days
t0 <- 0 

# started with a single infection
y0 <- c(N-1, 1, 0)

data <- list(N=N, cases=cases, n_days=n_days, ts=ts, t0=t0, y0=y0)
```

### DIFF
--- a/gui/src/app/pages/HomePage/DataGenerationWindow/DataGenerationWindow.tsx
+++ b/gui/src/app/pages/HomePage/DataGenerationWindow/DataGenerationWindow.tsx
@@ -38,22 +38,33 @@ const DataGenerationChildWindow: FunctionComponent<
   const consoleRef = useRef<HTMLDivElement | null>(null);
 
   const handleSetData = useCallback(
-    (data: any) => {
-      update({
-        type: "editFile",
-        content: JSON.stringify(data, null, 2),
-        filename: ProjectKnownFiles.DATAFILE,
-      });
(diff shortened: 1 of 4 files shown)


jsoules (approved):
Code changes are small and everything looks to be behaving as expected when loaded. I think this is probably good

magland:
Looks good to me.

WardBrian:
Made one final mostly-cosmetic change so that the "Data updated" message is only shown if the generated data differs from what was already there

WardBrian:
It says "Data unchanged"

I've just done one more update to change how print statements are captured -- what we were doing before wouldn't have worked for much beyond primitive types

jsoules (approved):
This makes sense (so far as I can follow webR), and comparing the live output between the head commit and the previous commit it looks to be a good change. I think you're good here.

Issue #149, opened 2024-07-24
data and analysis script examples available to user


magland:
I think it would be helpful for there to be a "?" icon at the top of the editors that would show documentation for writing the analysis.* and data.* scripts. Otherwise, I think a user wouldn't know how to get started when opening these tabs. I think this should be in the form of a markdown document that is opened as a popup window. I have some react components that can render markdown, including code snippets. But to start we could also just link to docs on github containing examples.

Let me know if you want me to get started on this.

WardBrian:
I think we should probably update the default values for these files to be a comment explaining them, rather than them defaulting to empty. That should be more than sufficient for the data ones, at least. Analysis scripts may need something closer to what you describe to fully explain them

magland:
> I think we should probably update the default values for these files to be a comment explaining them, rather than them defaulting to empty. That should be more than sufficient for the data ones, at least. Analysis scripts may need something closer to what you describe to fully explain them

I don't like the idea of every saved project having comment-only files called data.py, data.R, etc. I think they should be empty if not used, so they don't get included in the export

WardBrian:
That’s a fair point. We could update the export logic to check if they’re different from their defaults AND nonempty before including them, if we wanted. 

I think some users will honestly miss the ? — I hadn’t really taken note myself

magland:
How about, if the editor is empty, then display a special noticeable message in the toolbar: GET STARTED or something. If not empty, then it's just a "?" icon.

magland:
That's a cool idea. Let me take a crack at that for the stan editor.

magland:
... unless you already started.

PR #150, opened 2024-07-24
hint text in stan editor

magland:
If stan editor is empty, display hint message in the editor "Begin editing or select an example from the left panel" using monaco's addContentWidget.

And... no longer abuses syntax error window for this purpose.

See discussion in #149 

image

### DIFF
--- a/gui/src/app/FileEditor/TextEditor.tsx
+++ b/gui/src/app/FileEditor/TextEditor.tsx
@@ -27,6 +27,7 @@ type Props = {
   toolbarItems?: ToolbarItem[];
   label: string;
   codeMarkers?: CodeMarker[];
+  hintTextOnEmpty?: string;
 };
 
 export type ToolbarItem =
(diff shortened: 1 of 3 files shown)


gui/src/app/FileEditor/TextEditor.tsx
```diff
@@ -227,4 +242,24 @@ const NotSelectable: FunctionComponent<PropsWithChildren> = ({ children }) => {
...
 };
 
+const createHintTextContentWidget = (hintText: string) => {
+  return {
+    getDomNode: () => {
+      const node = document.createElement("div");
+      node.style.width = "max-content";
+      node.style.pointerEvents = "none";
+      node.textContent = hintText;
+      node.style.fontStyle = "italic";
```
WardBrian, line 252: Thoughts on pulling these `.style`s into a class? In particular, it would also be nice to make the text slightly lower contrast than it is currently, as vscode does

WardBrian (approved):
Overall looks great! One comment on styling, but otherwise should be good to go

gui/src/app/FileEditor/TextEditor.tsx
magland: I have created a css class EditorHintText for the italics and color. I feel like pointerEvents still belongs inline though.

PR #151, opened 2024-07-24
Create templates in script editors

magland:
This builds on #150 

In data.py, data.R, analysis.py... if the editor is empty, there is a LINK text hint inviting click to generate template code.

We'll want to think about what the template code should be, and whether we want to somehow link to more thorough documentation.

image

### DIFF
--- a/gui/src/app/FileEditor/TextEditor.tsx
+++ b/gui/src/app/FileEditor/TextEditor.tsx
@@ -108,7 +108,6 @@ const TextEditor: FunctionComponent<Props> = ({
     if (text || editedText) {
       return;
     }
-    console.log("--- contentOnEmpty:", contentOnEmpty);
     const contentWidget = createHintTextContentWidget(contentOnEmpty);
     editorInstance.addContentWidget(contentWidget);
     return () => {
--- a/gui/src/app/FileEditor/TextEditor.tsx
+++ b/gui/src/app/FileEditor/TextEditor.tsx
@@ -28,6 +28,7 @@ type Props = {
   label: string;
   codeMarkers?: CodeMarker[];
   hintTextOnEmpty?: string;
+  onClickHintText?: () => void;
 };
 
 export type ToolbarItem =
@@ -56,6 +57,7 @@ const TextEditor: FunctionComponent<Props> = ({
   label,
   codeMarkers,
   hintTextOnEmpty,
+  onClickHintText,
 }) => {
   const handleChange = useCallback(
     (value: string | undefined) => {
@@ -108,12 +110,15 @@ const TextEditor: FunctionComponent<Props> = ({
     if (text || editedText) {
       return;
     }
-    const contentWidget = createHintTextContentWidget(hintTextOnEmpty);
+    const contentWidget = createHintTextContentWidget(
+      hintTextOnEmpty,
+      onClickHintText,
+    );
     editorInstance.addContentWidget(contentWidget);
     return () => {
       editorInstance.removeContentWidget(contentWidget);
     };
-  }, [text, editorInstance, editedText, hintTextOnEmpty]);
+  }, [text, editorInstance, editedText, hintTextOnEmpty, onClickHintText]);
 
   /////////////////////////////////////////////////
 
@@ -242,14 +247,27 @@ const NotSelectable: FunctionComponent<PropsWithChildren> = ({ children }) => {
   return <div className="NotSelectable">{children}</div>;
 };
 
-const createHintTextContentWidget = (hintText: string) => {
+const createHintTextContentWidget = (
+  hintText: string,
+  onClick?: () => void,
+) => {
   return {
     getDomNode: () => {
       const node = document.createElement("div");
       node.style.width = "max-content";
-      node.style.pointerEvents = "none";
-      node.textContent = hintText;
       node.style.fontStyle = "italic";
+      if (onClick) {
+        const link = document.createElement("a");
+        link.textContent = hintText;
+        link.style.cursor = "pointer";
+        link.style.color = "gray";
+        link.style.textDecoration = "underline";
+        link.onclick = onClick;
+        node.appendChild(link);
+      } else {
+        node.style.pointerEvents = "none";
+        node.textContent = hintText;
+      }
       return node;
     },
     getId: () => "hintText",
--- a/gui/src/app/FileEditor/TextEditor.tsx
+++ b/gui/src/app/FileEditor/TextEditor.tsx
@@ -27,8 +27,7 @@ type Props = {
   toolbarItems?: ToolbarItem[];
   label: string;
   codeMarkers?: CodeMarker[];
-  hintTextOnEmpty?: string;
-  onClickHintText?: () => void;
+  contentOnEmpty?: string | HTMLSpanElement;
 };
 
 export type ToolbarItem =
@@ -56,8 +55,7 @@ const TextEditor: FunctionComponent<Props> = ({
   language,
   label,
   codeMarkers,
-  hintTextOnEmpty,
-  onClickHintText,
+  contentOnEmpty,
 }) => {
   const handleChange = useCallback(
     (value: string | undefined) => {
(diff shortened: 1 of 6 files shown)


gui/src/app/FileEditor/TextEditor.tsx
```diff
@@ -27,6 +27,8 @@ type Props = {
   toolbarItems?: ToolbarItem[];
   label: string;
   codeMarkers?: CodeMarker[];
+  hintTextOnEmpty?: string;
```
WardBrian, line 30: Would it be possible to have this be something like `string | HTMLSpanElement`?

In particular, I think it would be nice to have multiple links or be able to do what VSCode does an have some of the text be a link and some be click-through, e.g

image

Only the blue portion here is clickable.

This would let us make the text something like "Define a dictionary called `data` to update the data.json. [Click here]() to generate an example"

magland: I like that idea, I'll work on it, and make sure that the styling etc is taken care of inside the editor component.

magland: @WardBrian This is implemented in my latest commit.

gui/src/app/FileEditor/TextEditor.tsx
```diff
@@ -104,16 +104,17 @@ const TextEditor: FunctionComponent<Props> = ({
 
   useEffect(() => {
     if (!editorInstance) return;
-    if (!hintTextOnEmpty) return;
+    if (!contentOnEmpty) return;
     if (text || editedText) {
       return;
     }
-    const contentWidget = createHintTextContentWidget(hintTextOnEmpty);
+    console.log("--- contentOnEmpty:", contentOnEmpty);
```
jsoules, line 111: Do we want to leave this in for production, or is it a debug check?

gui/src/app/pages/HomePage/DataGenerationWindow/DataPyFileEditor.tsx
```diff
@@ -87,6 +87,21 @@ const DataPyFileEditor: FunctionComponent<Props> = ({
...
+    );
+    const a1 = document.createElement("a");
+    a1.onclick = () => {
+      setEditedFileContent(dataPyTemplate);
+    };
+    a1.textContent = "Click here to generate an example";
+    spanElement.appendChild(t1);
+    spanElement.appendChild(a1);
+    return spanElement;
+  }, [setEditedFileContent]);
```
jsoules, line 103: Is it possible to do this as a React component? No worries if not, it just seems like it'd be more straightforward than having to assemble it manually.

gui/src/app/pages/HomePage/DataGenerationWindow/DataRFileEditor.tsx
```diff
@@ -121,6 +121,21 @@ if (typeof(data) != "list") {
...
+    const a1 = document.createElement("a");
+    a1.onclick = () => {
+      setEditedFileContent(dataRTemplate);
+    };
+    a1.textContent = "Click here to generate an example";
+    spanElement.appendChild(t1);
+    spanElement.appendChild(a1);
+    return spanElement;
+  }, [setEditedFileContent]);
+
```
jsoules, line 138: Same question about whether we can make this a const component.

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
```diff
@@ -158,6 +158,21 @@ const AnalysisPyFileEditor: FunctionComponent<Props> = ({
...
+    const a1 = document.createElement("a");
+    a1.onclick = () => {
+      setEditedFileContent(analysisPyTemplate);
+    };
+    a1.textContent = "Click here to generate an example";
+    spanElement.appendChild(t1);
+    spanElement.appendChild(a1);
+    return spanElement;
+  }, [setEditedFileContent]);
+
```
jsoules, line 175: Same question

jsoules (commented):
Overall looks good, question about the hint-text content

gui/src/app/FileEditor/TextEditor.tsx
magland: oops. removed.

gui/src/app/pages/HomePage/DataGenerationWindow/DataPyFileEditor.tsx
magland: Would be nice, but I don't think it's possible.

gui/src/app/pages/HomePage/DataGenerationWindow/DataRFileEditor.tsx
```diff
@@ -121,6 +121,21 @@ if (typeof(data) != "list") {
     [fileContent, editedFileContent, handleRun, status, handleHelp],
   );
 
+  const contentOnEmpty = useMemo(() => {
+    const spanElement = document.createElement("span");
+    const t1 = document.createTextNode(
+      "Define a dictionary called data to update the data.json. ",
```
WardBrian, line 127:
```suggestion
      "Define a list called data to update the data.json. ",
```

R nomenclature suggestion

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
```diff
@@ -168,10 +183,27 @@ const AnalysisPyFileEditor: FunctionComponent<Props> = ({
       onSetEditedText={setEditedFileContent}
       readOnly={readOnly}
       toolbarItems={toolbarItems}
+      contentOnEmpty={contentOnEmpty}
     />
   );
 };
 
+const analysisPyTemplate = `import matplotlib.pyplot as plt
```
WardBrian, line 191: I think it is worth having `print(draws)` in this template as that also gives more help text.

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
```diff
@@ -168,10 +183,27 @@ const AnalysisPyFileEditor: FunctionComponent<Props> = ({
...
+
+# Show histogram for first parameter
+for pname in pnames:
+    print(pname)
+    samples = draws.get(pname)
+    print(samples.shape)
+    plt.hist(samples.ravel(), bins=30)
+    plt.title(pname)
+    plt.show()
+    break
```
WardBrian, line 204: I understand the intention, but I think it is confusing to write in a loop that breaks. The first parameter name will always be `lp__`, so maybe:

```python
# Get the parameter names
print(draws.parameter_names)

# plot the lp parameter
samples = draws.get("lp__")
print(samples.shape)
plt.hist(samples.ravel(), bins=30)
plt.title("lp__")
plt.show()

WardBrian (approved):
I think this is great! A few wording suggestions/suggestions about the template content.

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
magland: so that would need to be implemented on the Python class?

gui/src/app/pages/HomePage/DataGenerationWindow/DataRFileEditor.tsx
magland: Done

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
magland: Done

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
WardBrian: It is: https://github.com/flatironinstitute/stan-playground/blob/main/gui/src/app/pyodide/pyodideWorker/sp_load_draws.py#L37-L43

gui/src/app/pages/HomePage/DataGenerationWindow/DataPyFileEditor.tsx
WardBrian: I just looked into a few options, all the supported ways of turning a JSX element into a actual DOM element are in `react-dom/server` and therefore don't persist the `onClick` in a meaningful way. Too bad

gui/src/app/pyodide/AnalysisPyFileEditor.tsx
magland: Done

PR #154, opened 2024-07-25
Run data.R in fresh environment each time

WardBrian:
Without this, variables defined in the previous run are still available. It's the equivalent of `globals` in `pyodide`


magland:
Looks good.

PR #156, opened 2024-07-25
Update TabWidget to use 'overflow="auto" for scrolling

magland:
This change hides scrollbars when they are not needed. Here's what it looked like before the change. See lower-right corner.

image


WardBrian (approved):
Weird -- another rendering difference in Firefox. I think it just ignores overflow=scroll, or treats it as auto, because these weren't showing up for me

PR #158, opened 2024-07-26
project save/load to browser

magland:
Save projects to browser (uses IndexedDB)

Click "Save Project" button, then "Save to Browser"

It will be saved by the title of the project. If there already exists such a project, user will be prompted to replace.

Click "Load Project" button. Below the upload box the browser projects are listed. You can also click the delete icon to delete browser projects.


gui/src/app/pages/HomePage/SaveProjectWindow.tsx
```diff
@@ -211,4 +228,60 @@ const makeSPShareableLinkFromGistUrl = (gistUrl: string) => {
...
+    onClose();
+  }, [title, fileManifest, onClose]);
+
+  return (
+    <div className="SaveToBrowserView">
+      <h3>Save to Browser</h3>
+      <p>
+        This project will be saved to your browser as &quot;{title}&quot;.&nbsp;
+        It will be available to you on this device, but not on other devices or
+        browsers.
```
WardBrian, line 270: It should also probably be noted here that certain actions like clearing your browsers cache may lose this.

gui/src/app/pages/HomePage/BrowserProjectsInterface.ts
```diff
@@ -0,0 +1,108 @@
+export class BrowserProjectsInterface {
+  constructor(
+    private dbName: string = "stan-playground",
+    private storeName: string = "projects",
+  ) {}
+  async loadProject(title: string) {
+    const db = await this.openDatabase();
+    const transaction = db.transaction(this.storeName, "readonly");
+    const objectStore = transaction.objectStore(this.storeName);
```
jsoules, line 9: We  need to repeat these steps for basically any connection to the db (loading, saving, listing, or deleting) so might be nice to pull out into a function that just returns the `objectStore`.

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
```diff
@@ -15,6 +15,9 @@ import {
   useEffect,
   useState,
 } from "react";
+import { Hyperlink, SmallIconButton } from "@fi-sci/misc";
```
jsoules, line 18: I'd prefer `@mui.material.Link` over custom `Hyperlink` if possible.

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
```diff
@@ -211,4 +228,60 @@ const makeSPShareableLinkFromGistUrl = (gistUrl: string) => {
   return url;
 };
 
+type SaveToBrowserViewProps = {
+  fileManifest: Partial<FileRegistry>;
+  title: string;
+  onClose: () => void;
```
jsoules, line 234: I know we have this elsewhere in this file too--what's the intended use case? As far as I can see, it's just hard-coded to a no-op?

jsoules (commented):
A couple minor tweaks and questions.

I agree that we should give some warning that this mode of saving should be considered potentially volatile, since a cleared cache etc. would remove it.

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
```diff
@@ -103,9 +106,37 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
...
 
+  const [browserProjectTitles, setBrowserProjectTitles] = useState<string[]>(
+    [],
+  );
+  useEffect(() => {
+    const bpi = new BrowserProjectsInterface();
+    bpi.listProjects().then((titles) => {
+      setBrowserProjectTitles(titles);
+    });
+  }, []);
```
jsoules, line 117: Note that there is a race condition here if you have two tabs open: saving (or deleting) a project in one will not update the other.

I think this is probably okay, it's not really an intended operational mode or likely to come up, but I wanted to point it out.

gui/src/app/pages/HomePage/BrowserProjectsInterface.ts
magland: done

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
magland: switched to <Link> but I needed to style it so the cursor was pointer, for some reason.

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
magland: I changed the name to `onCancel` which should be clearer. This is called if user clicks the CANCEL button.

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
magland: I see what you mean, but this is not going to be a problem since retrieving from browser storage is fast enough.

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
magland: I updated this to read

```
This project will be saved to your browser as &quot;{title}&quot;.&nbsp;
        It will be available to you on this device until you clear your browser
        cache, but not on other devices or browsers.
```

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
jsoules: I don't think it's a matter of save/retrieve speed--it's if you have the window open already, there isn't anything triggering an update. (At least that's how it looked when I was playing around with it.)

Again, I think the steps to trigger are kind of far-fetched, so I'm not worried, but just want to note everything I see.

magland: Oh I see! Didn't read carefully enough :)

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
WardBrian: The [mui docs ](https://mui.com/material-ui/react-link/#accessibility)recommend using `component="button"` if there is no href, which is what the other things use

magland: I don't think letting the system style it as a button is appropriate here. By default the text becomes all caps, which is not suitable for titles which are case-sensitive keys.

WardBrian: That's not the behavior I observe with e.g. the retry Link in the connection window: https://github.com/flatironinstitute/stan-playground/blob/main/gui/src/app/CompilationServerConnectionControl/CompilationServerConnectionControl.tsx#L137

I believe component="button" doesn't change styling so much as it changes how the browser treats certain events in terms of accessibility and cursor behavior

magland: Got it. Made the change.

magland:
Is this okay to merge @WardBrian @jsoules ?

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
```diff
@@ -142,6 +174,45 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
...
+                </td>
+                <td>
+                  <Link
+                    onClick={() => {
+                      handleOpenBrowserProject(title);
+                    }}
+                    component="button"
+                    underline="none"
+                  >
+                    {title}
```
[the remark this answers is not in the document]
magland: I agree it would be helpful. But I'm wary about including it in meta.json as there's nothing that is enforcing the accuracy of that information if someone edits the project outside of SP. And for Gists, this information would be available independent of the content of meta.json.

Let me put the timestamp in the browser storage record and we'll see how that looks.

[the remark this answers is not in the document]
jsoules: I think this is a good idea (modulo all the usual complications about timestamps/time zones) but might be for a different PR? I'd want to see the other forms of persistence be aware of it also.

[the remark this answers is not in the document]
magland: (Didn't see your message Jeff until just now.)

I added timestamps for the saved browser projects, but it's not in meta.json.

The timestamps are now shown (in the form of "time-ago-strings", like "3 min ago").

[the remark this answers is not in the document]
magland: image

gui/src/app/util/timeAgoString.ts
```diff
@@ -0,0 +1,28 @@
...
+  const timestampSeconds = Math.floor(timestampMsec / 1000);
+  const now = Date.now();
+  const diff = now - timestampSeconds * 1000;
+  const diffSeconds = Math.floor(diff / 1000);
+  const diffMinutes = Math.floor(diffSeconds / 60);
+  const diffHours = Math.floor(diffMinutes / 60);
+  const diffDays = Math.floor(diffHours / 24);
+  const diffWeeks = Math.floor(diffDays / 7);
+  const diffMonths = Math.floor(diffWeeks / 4);
+  const diffYears = Math.floor(diffMonths / 12);
```
WardBrian, line 12: Since months is unused, diffWeeks/52 is probably a more accurate way to calculate years

magland: Okay I removed months, and am now using 365 days / yr. Not worrying about leap year for this.

magland:
This one is on hold for now because there are some questions about whether we want to have this functionality. Will revisit later.

PR #162, opened 2024-07-26
Update Stan syntax highlighting

WardBrian:
This defines a new tokenizer instead of relying on C++'s and a custom theme.

This also means we're not required to hard-code dark mode in.

image


jsoules (approved):
This looks good in the browser.

Beyond that, I don't feel qualified to judge the code proper, but I note that the changes are nowhere near as extensive as the diff would suggest.

magland:
I agree with Jeff

Issue #159, opened 2024-07-26
de-uglify the UI


magland:
* More consistent colors (and title bar color should not clash so much with the rest of the UI)
* Proper text alignment on left panel
* Less vertical space in top title bar
* Proper vertical text alignment on editor toolbars

WardBrian:
I’d like to add to this that we should probably have the text editors default to light UI, not the current dark style

magland:
I am going to do some work on this now. It certainly won't solve everything but I will aim to make an incremental improvement.

magland:
I made #167 #168 and #169

Now I will pause from working on this issue.

WardBrian:
With all of the above merged, is this issue completed?

magland:
> With all of the above merged, is this issue completed?

I think the remaining item is the consistent color scheme. We can open a separate issue for that.

PR #167, opened 2024-07-29
Update sidebar styling

magland:
Before / After

image

image

For #159


gui/src/app/pages/HomePage/Sidebar.tsx
```diff
@@ -62,7 +60,8 @@ const Sidebar: FunctionComponent<Sidebar> = ({
       {/* For spacing purposes */}
       <Toolbar />
 
-      <h3>Examples</h3>
+
+      <div className="SidebarContentWrapper SidebarHeading">Examples</div>
```
WardBrian, line 64: Is there a semantic reason to prefer (or not prefer) a styled div vs h3?

gui/src/app/pages/HomePage/Sidebar.tsx
```diff
@@ -72,43 +71,44 @@ const Sidebar: FunctionComponent<Sidebar> = ({
...
-              );
-              if (!ok) return;
-              update({ type: "clear" });
-            }}
-          >
-            Clear all
-          </Button>
-        </ListItem>
-      </List>
+      <div className="SidebarContentWrapper">
```
WardBrian, line 75: Similarly, it seems like actually rendering with `ul` (which is what the MUI ListItem renders as) might be more correct? If the margin is the issue we could place the entire list in this div

magland: That's okay with me if you want to make that. I'm happy with how it actually renders as is, and I think in the future we'll want to revamp the sidebar to accommodate more advanced functionality.

gui/src/app/pages/HomePage/Sidebar.tsx
jsoules: Technically there are accessibility reasons to prefer `H` tags where appropriate, but I didn't bring it up because a) we haven't been paying a lot of attention to that overall and b) it's not clear to me that it would actually be appropriate here anyway, as that sort of advice is really for a different sort of page/site...

magland:
@WardBrian @jsoules 

I went ahead and switched to using `<h3>` and `<ListItem>` again and updated the .css to achieve the margins, etc. that I was aiming for.

image

WardBrian (approved):

PR #170, opened 2024-07-29
Update toolbar, smarter hiding of labels

WardBrian:
This is an alternative to #168 

Prior:
image

Now:
image

Main differences are centering the items and some clever CSS to automatically hide the text labels when the container gets too small. This replaces the `useMediaQuery` hook in StanEditor, which didn't quite work anyway

### DIFF
--- a/gui/src/app/FileEditor/TextEditor.tsx
+++ b/gui/src/app/FileEditor/TextEditor.tsx
@@ -1,9 +1,8 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Editor, loader, useMonaco } from "@monaco-editor/react";
-import { Save } from "@mui/icons-material";
-import Button from "@mui/material/Button";
-import Link from "@mui/material/Link";
+
 import monacoAddStanLang from "@SpComponents/stanLang";
+import { ToolBar, ToolbarItem } from "@SpComponents/ToolBar";
 import { CodeMarker } from "@SpStanc/Linting";
 import { editor, MarkerSeverity } from "monaco-editor";
 import {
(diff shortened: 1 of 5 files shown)


magland:
Looks great!

gui/src/app/FileEditor/TextEditor.tsx
```diff
@@ -186,39 +173,95 @@ const toMonacoMarkerSeverity = (
...
+      });
+    }
+
+    if (readOnly) {
+      editorItems.push({
+        type: "text",
+        label: "Read Only",
+        color: "gray",
+      });
+    }
```
jsoules, line 218: Since the `readOnly` is logically incompatible with `edited`, this could be two simple conditionals:

```
if (readOnly) {
  push read only
} else {
  if (edited) {
    push edited stuff
  }
}
```

gui/src/app/FileEditor/TextEditor.tsx
```diff
@@ -186,39 +173,95 @@ const toMonacoMarkerSeverity = (
   }
 };
 
+type ToolbarProps = {
```
jsoules, line 176: If practical, I think it'd be neater for the `Toolbar` stuff to be in a separate file--it's less distracting compared to combining in the `TextEditor` component file imo.

jsoules (commented):
Couple suggestions. I think this is good.

I wonder if, when the icon labels are hidden, we can get a tighter spacing on the resulting icon-buttons; they seem like they stay pretty far apart?

WardBrian:
Icons are much tighter when collapsed now:

image

PR #171, opened 2024-07-29
update existing github gist

### LINKED ISSUE #153: feature: mechanism for updating gists
If a user is the owner of a gist and they update it (maybe even a small tweak like adding comments), it would be nice to allow them to be able to update the gist (create a revision) rather than creating a new gist. In order to allow this option in the "Save Project" dialog we would somehow need to keep track of what the source gist was for the project. I don't think storing that in the meta.json of the project is the right thing to do -- maybe there's another way.

magland:
In save project window, user can either save to new gist or update existing gist. When updating the existing Gist, they need to manually enter the Gist URL (`https://gist.github.com/<user>/<id>`). For this PR we don't keep track of the origin gist URL for the open project (even though it would be convenient) because there are a number of questions about how that would be done. I think that should be a separate PR.

Closes #153


gui/src/app/gists/saveAsGitHubGist.ts
```diff
@@ -67,6 +67,7 @@ export const updateGitHubGist = async (
       files[path] = { content };
     }
   }
+  console.log("--------------- files", files);
```
jsoules, line 70: Is this intended to be in the production version?

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
```diff
@@ -204,6 +283,52 @@ const GistExportView: FunctionComponent<GistExportViewProps> = ({
...
+  }
+  // handle deleted files
+  for (const fname in existingFiles) {
+    if (!newFiles[fname] || !newFiles[fname].trim()) {
+      patch[fname] = null;
+    }
+  }
+  return patch;
+};
+
```
jsoules, line 331: Gist-manipulation logic should go in the `gists/SaveAsGithubGist` rather than in the UI component

jsoules (changes requested):

gui/src/app/gists/saveAsGitHubGist.ts
magland: fixed

gui/src/app/pages/HomePage/SaveProjectWindow.tsx
magland: Done

magland:
This is going to have a conflict with #158, so after one of these is merged I'll work on resolving it.

magland:
I'm going to need to do increasingly complex conflict resolution for this and #158 once the latest PRs are merged. Is there a reason why these two are not getting merged? I just would like to know when I should incorporate those changes.

@WardBrian @jsoules

WardBrian:
I don't have any issue with the functionality of this PR, but I do find the logic to create the `GistExplainer` div to be a bit hard to follow now

I haven't looked at #158 in a few days. I'm not entirely convinced it's a good functionality to provide (it seems almost certain to me that people will misunderstand it and lose their data), but I'm not so opposed as to prevent it being included

magland:
Okay I'll work on resolving conflicts for this one first, and try to clarify/simplify the GistExplainer

magland:
I've resolved the conflicts here. @WardBrian could you elaborate a bit on where you are confused? Are you talking about this code?

https://github.com/flatironinstitute/stan-playground/blob/8476c9652453341228aa2b684d8843d51c07ec0f/gui/src/app/pages/HomePage/SaveProjectWindow.tsx#L51-L94

WardBrian:
No, it's really more the code on line 169-on that actually builds up the view -- I understand the desire to share between the uploading and updating case, but it leads to a lot of conditionals and even conditional prose. 

I think it would make sense to split them, because a user who is updating an existing gist has already been through the token creation, presumably?

magland:
> No, it's really more the code on line 169-on that actually builds up the view -- I understand the desire to share between the uploading and updating case, but it leads to a lot of conditionals and even conditional prose.
> 
> I think it would make sense to split them, because a user who is updating an existing gist has already been through the token creation, presumably?

Okay I'll take a crack at that.

jsoules (approved):
I'm okay moving forward with this.

magland:
@WardBrian I made a separate component for updating the gist, and tried to reuse components where it seemed appropriate.

WardBrian:
Thanks -- looks great to me!

WardBrian (approved):

PR #172, opened 2024-07-29
Hide minimap on editor

WardBrian:
This primarily removes the "minimap" sidebar monaco provides:
image

This is not really necessary in our project and has always led to odd flickering with the splitters


magland:
Looks good

PR #175, opened 2024-07-30
webR: Support streaming output

WardBrian:
This uses some examples from the webR documentation to produce something that looks a lot more like the way we handle output in Python. In particular, outputs (both text and plots) will populate while the script is running, rather than at the end. 

Because webR provides this via a async API, I had to use their pattern of a infinite (or at least not obviously terminating) async loop, which I find kind of strange. Getting this right required a bit of extra state management, but the end result is pretty nice to use.


magland:
I tried it out on my machine and it's very smooth for the analysis.R example that we've been working with. It's really helpful to show the console progress during package loading. 

The usage also looks nice.

Not sure how much you want me to review that new file. I believe it's doing the right thing, but I can take a closer look if you want.

WardBrian:
> Not sure how much you want me to review that new file. I believe it's doing the right thing, but I can take a closer look if you want.

The only part I feel somewhat uncertain of is the async loop, but I think it is working, and it is based on their examples pretty heavily. 

Also, do you know if the useMemos I put in the call sites are actually valuable?

magland:
> The only part I feel somewhat uncertain of is the async loop, but I think it is working, and it is based on their examples pretty heavily.

The only thing I would suggest is to have a way to terminate that infinite loop when the useEffect gets closed. That would require passing in something like `const closedHandle = {closed}`, and then setting `closedHandle.closed = true` when exiting the useEffect. Then checking for `closedHandle.closed` in the loop.

> Also, do you know if the useMemos I put in the call sites are actually valuable?

In this case, useMemo is not needed since you are only using the individual components in the hook dependencies in useWebR. I think you should pass these arguments directly to useWebR() rather than making a separate variable webRArgs.

gui/src/app/Scripting/Analysis/AnalysisRWindow.tsx
```diff
@@ -28,19 +28,14 @@ const AnalysisRWindow: FunctionComponent<AnalysisWindowProps> = ({
     notRunnableReason,
   } = useAnalysisState(latestRun);
 
+  const { run } = useWebR({ consoleRef, imagesRef, onStatus });
```
jsoules, line 31: Looking inside the custom hook, Jeremy is right that you always depend on the individual components of the object rather than the object itself. So this should be fine not memoized. But I'd want to double-check just to be sure.

magland: As long as consoleRef, imagesRef, onStatus are being received by the hook individually, then you definitely do not need to memoize the args object. It's just a way of passing in individual args.

magland:
Changes look good to me.

gui/src/app/Scripting/webR/useWebR.ts
```diff
@@ -0,0 +1,189 @@
...
+          }
+
+          const env = await new shelter.REnvironment(globals);
+
+          const options = { ...captureOutputOptions, env };
+
+          // setup
+          await webR.evalRVoid(
+            `webr::shim_install()
+webr::canvas()`,
```
jsoules, line 89: These might be a hair more legible if you throw them into a const somewhere with less indentation & give them a name.

gui/src/app/Scripting/webR/useWebR.ts
```diff
@@ -0,0 +1,189 @@
...
+          await webR.evalRVoid(code, options);
+
+          if (onData) {
+            const result = JSON.parse(
+              await webR.evalRString(
+                `if (typeof(data) != "list") {
+stop("[stan-playground] data must be a list")
+}
+.SP_DATA <- jsonlite::toJSON(data, pretty = TRUE, auto_unbox = TRUE)
+invisible(.SP_DATA)`,
```
jsoules, line 101: These might be a hair more legible if you throw them into a const somewhere with less indentation & give them a name.

jsoules (approved):
I think I get it & it looks reasonable to me.

WardBrian:
@jsoules I used the trick we've used elsewhere to put them in actual .R files so we even get nicer syntax highlighting in our editor now

PR #176, opened 2024-07-30
Support Ctrl+Enter to run scripts

WardBrian:
TextEditor takes a generic list of [`IActionDescriptors`](https://microsoft.github.io/monaco-editor/typedoc/interfaces/editor.IActionDescriptor.html), and ScriptEditor provides one that runs the current script


magland:
Confirmed that it works. Looks good.

jsoules (approved):
Seems reasonable.

jsoules:
Yup, good catch to memoize the lists.

PR #177, opened 2024-07-30
Confirm before unload if unsaved changes

### LINKED ISSUE #173: “Are you sure” before exiting or re-loading with dirty state
I think we discussed this at some point. It would obviously have been pretty annoying during development, but for real users this seems valuable

WardBrian:
Closes #173


jsoules (approved):
Note that the expected behavior is to cause a standard popup for navigation events away from the page (incl. closing or refreshing the tab), but does not affect loading an example, etc.

I think this does what it intends to do and should be included.

PR #180, opened 2024-07-31
Use MUI for connection controls

WardBrian:
Follow on to #178, this addresses the couple of my comments that were deferred

image


gui/src/app/CompilationServerConnectionControl/ConfigureCompilationServerDialog.tsx
```diff
@@ -20,103 +29,103 @@ const ConfigureCompilationServerDialog: FunctionComponent<
...
+          <p>
+            The public server is provided for convenience, but may not be as
+            reliable as a local server, depending on the current load and
+            availability.
+          </p>
+        </div>
+      )}
+
+      <Divider />
+      <Button onClick={() => closeDialog()}>Close</Button>
```
jsoules, line 128: Wonder if we really need this, given that there's the `X` and it'll close if you click outside the dialog?

jsoules (approved):
Two other thoughts to consider:

- There's some weird spacing issues with the "retry" link: vertically, because of the hidden underline, and horizontally, it falls a little too close to the connection status text. I'd consider bumping that out, and possibly just using a `Button` instead of a `Link`.
- If possible, it might be nice to find a way to display the URLs for the public/local connection compilation server. Not a deal-breaker.

I'd be happy to merge this even if you didn't address those, however.

gui/src/app/CompilationServerConnectionControl/ConfigureCompilationServerDialog.tsx
WardBrian: Agreed, removed

WardBrian:
Updated screenshot:
image

PR #181, opened 2024-07-31
Allow multiple files in upload window, give more control

### LINKED ISSUE #95: Uploading a single file should not hide the drop area
I think uploading a single file should keep the drop box enabled to allow you to also add a second/third etc. Right now if you want to upload multiple files but do not drag them all at the same time, you need to click import again and then select the 'add to existing' option

_Originally posted by @WardBrian in https://github.com/flatironinstitute/stan-playground/pull/80#discussion_r1655135279_

WardBrian:
Closes #95 

You can now drop multiple items in succession, and you are shown a list with the ability to remove items before clicking the create project buttons. 

image

### DIFF
--- a/gui/src/app/pages/HomePage/LoadProjectWindow.tsx
+++ b/gui/src/app/pages/HomePage/LoadProjectWindow.tsx
@@ -34,8 +34,9 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
         files: fileManifest,
         clearExisting: true,
       });
+      onClose();
     },
-    [update],
+    [onClose, update],
   );
 
   const importUploadedFiles = useCallback(
--- a/gui/src/app/pages/HomePage/LoadProjectWindow.tsx
+++ b/gui/src/app/pages/HomePage/LoadProjectWindow.tsx
@@ -28,13 +28,17 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
 
   const importUploadedZip = useCallback(
     async (zipFile: ArrayBuffer) => {
-      const fileManifest = await deserializeZipToFiles(zipFile);
-      update({
-        type: "loadFiles",
-        files: fileManifest,
-        clearExisting: true,
-      });
-      onClose();
+      try {
+        const fileManifest = await deserializeZipToFiles(zipFile);
+        update({
+          type: "loadFiles",
+          files: fileManifest,
+          clearExisting: true,
+        });
+        onClose();
+      } catch (e: any) {
+        setErrorText(e.message);
+      }
     },
     [onClose, update],
   );
--- a/gui/src/app/pages/HomePage/LoadProjectWindow.tsx
+++ b/gui/src/app/pages/HomePage/LoadProjectWindow.tsx
@@ -1,6 +1,8 @@
+import { Delete } from "@mui/icons-material";
 import Button from "@mui/material/Button";
+import IconButton from "@mui/material/IconButton";
+import Stack from "@mui/material/Stack";
 import {
-  FieldsContentsMap,
   FileNames,
   FileRegistry,
   mapFileContentsToModel,
@@ -8,13 +10,9 @@ import {
 import { ProjectContext } from "@SpCore/ProjectContextProvider";
 import { deserializeZipToFiles, parseFile } from "@SpCore/ProjectSerialization";
 import UploadFilesArea from "@SpPages/UploadFilesArea";
-import {
-  FunctionComponent,
-  useCallback,
-  useContext,
-  useEffect,
-  useState,
-} from "react";
+import { FunctionComponent, useCallback, useContext, useState } from "react";
+
+type File = { name: string; content: ArrayBuffer };
 
 type LoadProjectWindowProps = {
   onClose: () => void;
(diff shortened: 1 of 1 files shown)


gui/src/app/pages/HomePage/LoadProjectWindow.tsx
```diff
@@ -89,58 +81,85 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
...
-      // replace the current project.
-      setShowReplaceProjectOptions(true);
-    } else {
-      // Otherwise, we just go ahead and import the files, replacing the
-      // entire project
-      importUploadedFiles({ replaceProject: true });
-    }
-  }, [filesUploaded, importUploadedFiles]);
+  const onUpload = useCallback(
+    (fs: { name: string; content: ArrayBuffer }[]) => {
```
jsoules, line 85: Can use `File` type alias defined above

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
```diff
@@ -89,58 +81,85 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
...
+                </tbody>
+              </table>
+            </div>
+            <div>
+              <Button
+                onClick={() => importUploadedFiles({ replaceProject: true })}
+              >
+                Load into a NEW project
+              </Button>
+              <span style={{ margin: "0 10px" }}>or</span>
```
jsoules, line 153: Embedded const styling

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
```diff
@@ -89,58 +81,85 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
...
-      // Otherwise, we just go ahead and import the files, replacing the
-      // entire project
-      importUploadedFiles({ replaceProject: true });
-    }
-  }, [filesUploaded, importUploadedFiles]);
+  const onUpload = useCallback(
+    (fs: { name: string; content: ArrayBuffer }[]) => {
+      if (fs.length === 1 && fs[0].name.endsWith(".zip")) {
+        setFilesUploaded(fs);
+        importUploadedFiles({ replaceProject: true });
```
magland, line 88: This is not going to work. The setFilesUploaded(fs) is not going to take effect until the next render, so importUploadedFiles() should not be called immediately.

What I would suggest  is to have a separate useEffect that checks for the case of a single .zip file and then calls the importFilesUploaded()

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
```diff
@@ -49,25 +43,19 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
...
+            if (file.name.endsWith(".stan")) {
+              if (stanFileName !== "") {
+                throw Error("Only one .stan file can be uploaded at a time");
+              }
+              files["main.stan"] = parseFile(file.content);
+              stanFileName = file.name;
+              continue;
+            }
             if (!Object.values(FileNames).includes(file.name as any)) {
               throw Error(`Unrecognized file: ${file.name}`);
```
magland, line 60: Suppose the user uploaded multiple files including something.zip. This message would then say "Unrecognized file: something.zip" whereas it should really say: "When importing from a Zip file, you must only upload the one .zip file" or something to that effect.

Also, I'm not sure "Unrecognized file" is the right message here. Maybe, "Unsupported file name"

magland (changes requested):

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
```diff
@@ -89,58 +81,85 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
...
+          You can upload:
+          <ul>
+            <li>A .zip file that was previously exported</li>
+            <li>
+              A directory of files that were extracted from an exported .zip
+              file
+            </li>
+            <li>An individual *.stan file</li>
+            <li>
+              Other individual project files (meta.json, data.json, init.json,
```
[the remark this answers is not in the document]
WardBrian: Nope, oversight by me.

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
WardBrian: Updated to make both changes

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
WardBrian: Yep, should have tested this! I ended up splitting up the zip logic from the function that gets called when the buttons are clicked

gui/src/app/pages/HomePage/LoadProjectWindow.tsx
```diff
@@ -24,63 +23,64 @@ const LoadProjectWindow: FunctionComponent<LoadProjectWindowProps> = ({
...
+        }
+
+        const fileManifest = mapFileContentsToModel(files);
+        update({
+          type: "loadFiles",
+          files: fileManifest,
+          clearExisting: replaceProject,
+        });
+
+        if (stanFileName !== "" && fileManifest.meta === undefined) {
```
jsoules, line 80: This is supposed to retitle the project if the .stan file uploaded is not called main.stan, right? If so I'm not sure that's working...

WardBrian: It appears to be working for me? Though it should probably only do this if you hit 'new project'

jsoules: Ah, it only functions if you don't have a meta.json in the upload. My mistake.

WardBrian: Yeah, we let the title in `meta.json` be the definitive one in the case that there are both

magland (approved):

PR #182, opened 2024-07-31
Minor visual tweaks

WardBrian:
- When the sampler fails, the current output is pretty messy. This adds some light formatting. The easiest way to observe this is delete a variable from data.json
- A few more MUI button element changes, and a bit of extra space in the "draws" output tab


jsoules (approved):
Looks good. Confirmed that the sampler failure output is nicely formatted and that the sample downloads still work.

Issue #161, opened 2024-07-26
add compile button next to "stan model not compiled"


magland:
Should we add an additional compile button next to the "stan model not compiled" text?

PR #184, opened 2024-07-31
reduce redundant compile jobs, cache main.js URL

### LINKED ISSUE #138: Reduce redundant compilation jobs
Right now every time you reload the page a new compilation job is created. It uses the cached wasm file, but it does create a new directory, etc. To reduce redundancy I propose:

In browser local storage, keep a mapping (stan file hash) -> job ID. So when it comes time to compile, the client checks this table and checks for the existence of the wasm file. If it exists, then it uses that. If not, then it removes the entry in the table and proceeds as before.

magland:
Closes #138 

This replaces the existing compile cache mechanism which is not ideal (see below).

In this PR, we maintain a simple cache in local storage that maps stan program hashes to main.js URLs (main.js is created during compilation). Each time the main.stan is saved (or on initial load), we check this cache to see if the URL exists (meaning that a job had already completed). On cache hit we still need to verify that the remote file exists (maybe the compilation server was restarted). If not exists, clear that entry from the cache. If it does exist, then we don't need to create a new job, we'll just use that. If this cache (which should be very small) ever gets to more than 50 entries, it is simply cleared to avoid out-of-space error.

EXISTING (non-ideal) functionality

Every time the page is reloaded, a new compilation job is created. There is a localStorage cache but it doesn't point to the main.js URL, but just keeps track about whether the program has been compiled. There are a number of problems with this strategy (I know, I coded it in the first place), and I won't take the time to enumerate all the negatives (unless prompted).

### DIFF
--- a/gui/src/app/FileEditor/StanFileEditor.tsx
+++ b/gui/src/app/FileEditor/StanFileEditor.tsx
@@ -5,7 +5,6 @@ import TextEditor from "@SpComponents/TextEditor";
 import { ToolbarItem } from "@SpComponents/ToolBar";
 import compileStanProgram from "@SpStanc/compileStanProgram";
 import { stancErrorsToCodeMarkers } from "@SpStanc/Linting";
-import { checkMainJsUrlCache } from "@SpStanc/mainJsUrlCache";
 import useStanc from "@SpStanc/useStanc";
 import {
   FunctionComponent,
@@ -97,16 +96,6 @@ const StanFileEditor: FunctionComponent<Props> = ({
     compileStatus,
     setCompiledUrl,
   ]);
-
-  useEffect(() => {
-    // if the stan program has already been compiled, and the record of that is in the cache,
-    // then call handleCompile, which will end up loading from the cache
-    checkMainJsUrlCache(fileContent).then((mainJsUrl) => {
-      if (mainJsUrl) {
-        handleCompile();
-      }
-    });
-  }, [fileContent, handleCompile]);
 
   const [syntaxWindowVisible, setSyntaxWindowVisible] = useState(false);
 
--- a/gui/src/app/FileEditor/StanFileEditor.tsx
+++ b/gui/src/app/FileEditor/StanFileEditor.tsx
@@ -5,6 +5,7 @@ import TextEditor from "@SpComponents/TextEditor";
 import { ToolbarItem } from "@SpComponents/ToolBar";
 import compileStanProgram from "@SpStanc/compileStanProgram";
 import { stancErrorsToCodeMarkers } from "@SpStanc/Linting";
+import { checkMainJsUrlCache } from "@SpStanc/mainJsUrlCache";
 import useStanc from "@SpStanc/useStanc";
 import {
   FunctionComponent,
@@ -79,16 +80,6 @@ const StanFileEditor: FunctionComponent<Props> = ({
     setCompiledUrl(mainJsUrl);
     setCompileStatus("compiled");
     setTheStanFileContentThasHasBeenCompiled(fileContent);
-
-    // record in local storage that we compiled this particular stan file
-    try {
-      const key = getKeyNameForCompiledFile(stanWasmServerUrl, fileContent);
-      const value = JSON.stringify({ mainJsUrl });
-      localStorage.setItem(key, value);
-    } catch (e: any) {
-      console.error("Problem recording compiled file in local storage");
-      console.error(e);
-    }
   }, [fileContent, setCompiledUrl]);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


gui/src/app/FileEditor/StanFileEditor.tsx
```diff
@@ -107,21 +98,15 @@ const StanFileEditor: FunctionComponent<Props> = ({
     setCompiledUrl,
   ]);
 
-  const [didInitialCompile, setDidInitialCompile] = useState(false);
```
WardBrian, line 110: It seems like it might be valuable to keep something like this? I think as this is currently implemented, we're doing a fetch every time a file is saved, not just on initial load?

gui/src/app/compileStanProgram/mainJsUrlCache.ts
```diff
@@ -0,0 +1,69 @@
+export const checkMainJsUrlCache = async (
+  stanProgram: string,
+): Promise<string | null> => {
+  const mainJsCache = localStorage.getItem("mainJsCache");
+  if (!mainJsCache) return null;
+  try {
+    const cache = JSON.parse(mainJsCache);
```
WardBrian, line 7: Is there a solution that doesn't involve parsing/writing to JSON so frequently? I've read a bit about react's `useSyncExternalStore` hook which seems like it might be for this kind of thing?

WardBrian:
What is this hoping to reduce, exactly? It seems like it is probably making the same number of web requests, and given that the compilations are themselves cached on the backend I would probably vote for removing this local storage trick entirely rather than expanding it

gui/src/app/FileEditor/StanFileEditor.tsx
magland: No, on each save it does not do a remote fetch unless it found the thing in the cache and needs to verify that it still exists on the server. Reading this small item from localStorage is very inexpensive.

WardBrian: Ah, right. Overlooked that

magland:
> What is this hoping to reduce, exactly? It seems like it is probably making the same number of web requests, and given that the compilations are themselves cached on the backend I would probably vote for removing this local storage trick entirely rather than expanding it

It's not the same number of web requests - this actually saves quite a lot. With existing system, even though the compilation has already completed, a new job is still initiated, the stan program is still uploaded, and other requests are still made.

I don't think we want all this to happen every time the user refreshes the page or loads the same example.

gui/src/app/compileStanProgram/mainJsUrlCache.ts
magland: We can consider it I suppose. But I want to emphasize that this is going to be a rather small string. A lot more expensive things are happening each time user saves a file.

WardBrian:
I think my preference would be _zero_ automatic requests. Just wait until the user clicks "Compile" (especially if we implement #161), and then let the server cache do its business. 

Especially if we imagine users switching between different servers I think this gets messy (as currently implemented, will this cause the client to send a request to a server other than the currently selected one in the UI?)

magland:
> I think my preference would be _zero_ automatic requests. Just wait until the user clicks "Compile" (especially if we implement #161), and then let the server cache do its business.
> 
> Especially if we imagine users switching between different servers I think this gets messy (as currently implemented, will this cause the client to send a request to a server other than the currently selected one in the UI?)

Okay, I have removed the automatic compile on save... and I now we do not use cache if it is for a compile server other than the one selected.

I think it is still valuable to have this browser cache to (dramatically) reduce the number of job directories being created on the server and to speed up the UI experience. Try it out and look at the network activity to appreciate the difference.

WardBrian:
The latest set of changes looks good to me. I haven't had the chance to try it out, but it seems like a nice thing to be doing.

Your comment reminded me we should probably be doing something on the compilation server to garbage collect job directories over time.

gui/src/app/compileStanProgram/mainJsUrlCache.ts
```diff
@@ -0,0 +1,74 @@
...
+    const url = cache[stanProgramHash];
+    if (!url) return null;
+    if (!url.startsWith(baseStanWasmServerUrl)) {
+      // the url is not from the current server
+      return null;
+    }
+    // check to see if the url is still valid
+    const exists = await checkRemoteFileExists(url);
+    if (!exists) {
+      console.warn("mainJsCache url no longer exists");
```
WardBrian, line 19: Not sure if this is worth warning over -- we expect it to happen when the server restarts, etc

WardBrian (approved):
Tried it out -- one minor comment, otherwise looks good!

gui/src/app/compileStanProgram/mainJsUrlCache.ts
magland: removed.

PR #189, opened 2024-07-31
Move server choice state up for better consistency

WardBrian:
Follow on to #180, this moves the state up a bit which gives us a more consistent behavior when you refresh the page and then click on the control pop-up (previously, the radio button would always initialize to "Public", even if that is not what the page was actually storing).


magland:
Looks good to me

PR #191, opened 2024-07-31
Compile Model button and CompileContextProvider

### LINKED ISSUE #161: add compile button next to "stan model not compiled"

magland:
Closes #161

Added a big "Compile Model" button in the sampling area. More intuitive for the user I believe, so they don't need to read a message "stan model not compiled" and then search for the small button above the editor.

In order to do this I needed to lift the compilation state way up... so I created a CompileContext and moved the handleCompile callback hook into CompileContextProvider. This simplified some things - for example, compiledMainJsUrl doesn't need to be passed around in props.

The one awkward thing was needing to pass the validSyntax boolean up to the context from the editor. That's necessary so that the compile button knows when it needs to be disabled. Aside from that, I think everything has become tidier, IMO.

Once the user clicks compile, their mouse is right there for the "Run Sampling" press, which is nice especially when the compilation has been cached.

image


gui/src/app/CompileContext/CompileContextProvider.tsx
```diff
@@ -0,0 +1,86 @@
+import { ProjectContext } from "@SpCore/ProjectContextProvider";
+import compileStanProgram from "@SpStanc/compileStanProgram";
```
WardBrian, line 2: thoughts on moving the `compileStanProgram.ts` file into this folder?

gui/src/app/CompileContext/CompileContextProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const handleCompile = useCallback(async () => {
+    setCompileStatus("compiling");
+    await new Promise((resolve) => setTimeout(resolve, 500)); // for effect
+    const onStatus = (msg: string) => {
+      setCompileMessage(msg);
+    };
+    const stanWasmServerUrl =
+      localStorage.getItem("stanWasmServerUrl") ||
+      "https://trom-stan-wasm-server.magland.org";
```
WardBrian, line 56: This is relying on the fact that the compilation control sets local storage, but should this setting just live in this context rather than in that UI?

gui/src/app/FileEditor/StanFileEditor.tsx
```diff
@@ -22,81 +22,38 @@ type Props = {
...
 }) => {
   const { stancErrors, requestFormat } = useStanc(
     "main.stan",
     editedFileContent,
     setEditedFileContent,
   );
 
+  const { setValidSyntax } = useContext(CompileContext);
+
+  const { compileStatus, compileMessage, compile } = useContext(CompileContext);
```
WardBrian, line 43: I think these two can be combined?

gui/src/app/FileEditor/StanFileEditor.tsx
```diff
@@ -22,81 +22,38 @@ type Props = {
...
+
   const validSyntax = useMemo(() => {
     return stancErrors.errors === undefined;
   }, [stancErrors]);
 
+  useEffect(() => {
+    setValidSyntax(validSyntax);
+  }, [validSyntax, setValidSyntax]);
+
```
WardBrian, line 52: [optional, not sure if it's actually a good idea] this logic could be moved inside the useStanc hook

gui/src/app/RunPanel/RunPanel.tsx
```diff
@@ -39,62 +45,90 @@ const RunPanel: FunctionComponent<RunPanelProps> = ({
...
-            variant="contained"
-            color="success"
-            onClick={handleRun}
-            disabled={runStatus === "sampling" || runStatus === "loading"}
-          >
-            run sampling
-          </Button>
-          &nbsp;
-          {runStatus === "sampling" && (
+        {compileStatus === "compiled" ? (
```
WardBrian, line 59: This now has enough `? :` and `&&`s in it that I find it pretty difficult to follow. Do you think we could split this up either into a few small components or something like a `content` variable with some if statements before the return?

WardBrian (commented):
This looks pretty good to me! A few questions and suggestions

gui/src/app/CompileContext/CompileContextProvider.tsx
magland: With https://github.com/flatironinstitute/stan-playground/pull/191/commits/675b37f07b9d7fe344b20796e4dba9dc8c7deb70 I moved stanWasmServerUrl into CompileContext and it simplified a lot, and the interaction with stanWasmServerUrl in localStorage is now in one place. In that commit I also fixed some issues with the compile server config dialog - it seems to be working smoothly now.

gui/src/app/CompileContext/CompileContextProvider.tsx
magland: Done

gui/src/app/CompileContext/CompileContextProvider.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  useEffect(() => {
+    // if the compiled content is not the same as the current content,
+    // then the state should not be compiled or failed
+    if (data.stanFileContent !== theStanFileContentThasHasBeenCompiled) {
+      if (compileStatus === "compiled" || compileStatus === "failed") {
+        setCompileStatus("");
+        setCompiledMainJsUrl("");
+      }
+    }
```
jsoules, line 44: Is it worthwhile to do the flip side of this--such that if the `stanFileContent` is flipped back to equal the content that has been compiled, we revert to the "compiled" state?

I'm imagining deleting and re-adding a semicolon for instance.

gui/src/app/CompileContext/CompileContextProvider.tsx
magland: I realized that some of those things had already been fixed in main branch and there were some conflicts. Resolved the conflicts.

gui/src/app/FileEditor/StanFileEditor.tsx
magland: Can we figure this out later?

gui/src/app/CompileContext/CompileContextProvider.tsx
```diff
@@ -0,0 +1,102 @@
...
+    setCompiledMainJsUrl,
+  ]);
+
+  const [stanWasmServerUrl, setStanWasmServerUrl] = useState<string>(
+    initialStanWasmServerUrl,
+  );
+  useEffect(() => {
+    // persist to local storage
+    localStorage.setItem("stanWasmServerUrl", stanWasmServerUrl);
+  }, [stanWasmServerUrl]);
```
jsoules, line 58: Not sure I understand why we need to persist this to local storage vs. it being part of the context?

gui/src/app/FileEditor/StanFileEditor.tsx
magland: Done

gui/src/app/RunPanel/RunPanel.tsx
magland: Done

magland:
Resolved all conflicts and tested this.

gui/src/app/RunPanel/RunPanel.tsx
```diff
@@ -39,62 +45,94 @@ const RunPanel: FunctionComponent<RunPanelProps> = ({
...
+        <Button
+          variant="contained"
+          onClick={compile}
+          disabled={isCompileModelDisabled(projectData, validSyntax)}
+        >
+          compile model
+        </Button>
       </div>
+    );
+  }
```
jsoules, line 117: Even pulled out of the return statement itself, this is still pretty long & tough to follow; the branch where the Stan code has successfully been compiled, in particular, is quite a mouthful.

I think I'd be happier if the 3 branches of the conditional were placed in components & then the conditional just chose from the 3 components.

Actually, I think I'd want to put the entire `compileStatus==="compiled"` branch in a separate file, and have that return 3-4 individual versions based on the compile status--but having the conditional at the top rather than running throughout the single `div` element.

Let me know if you'd prefer that I just do this--I know it can be frustrating to have someone ask for something fairly specific without giving enough details.

jsoules (commented):
A few suggestions. Feel like I need to think some more, but I definitely see that having the compilation status as a context simplifies quite a bit of the state-passing.

gui/src/app/CompileContext/CompileContextProvider.tsx
magland: I'd vote against that because (a) it would be challenging to manage that -- you see in this code the compiledMainJsUrl gets cleared (b) it's easy enough to click the compile button again, and it will be a cache hit.

gui/src/app/CompileContext/CompileContextProvider.tsx
magland: When user refreshes the page, or closes the tab and comes back later, we want this to remember the choice.

gui/src/app/RunPanel/RunPanel.tsx
magland: Sure, if you want to do that would be helpful. If I understand what you are proposing, I agree with it.

gui/src/app/FileEditor/StanFileEditor.tsx
WardBrian: Fine by me!

gui/src/app/CompileContext/CompileContextProvider.tsx
WardBrian: (this is also the current behavior, the code just moved in this PR)

gui/src/app/RunPanel/RunPanel.tsx
jsoules: If you don't mind waiting, give me 10-15 and I'll put in a PR against this branch.

gui/src/app/RunPanel/RunPanel.tsx
```diff
@@ -39,62 +45,94 @@ const RunPanel: FunctionComponent<RunPanelProps> = ({
...
   );
 };
 
+const isCompileModelDisabled = (
+  projectData: ProjectDataModel,
+  validSyntax: boolean,
+) => {
+  if (!projectData.stanFileContent.trim()) return true;
+  if (modelHasUnsavedChanges(projectData)) return true;
+  if (!validSyntax) return true;
```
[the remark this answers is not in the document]
magland: Done

magland:
Good to merge?

gui/src/app/CompilationServerConnectionControl/ConfigureCompilationServerDialog.tsx
```diff
@@ -1,38 +1,36 @@
...
+> = ({ isConnected, onRetry }) => {
+  const { stanWasmServerUrl, setStanWasmServerUrl } =
+    useContext(CompileContext);
+
+  const serverType: ServerType =
+    stanWasmServerUrl === publicUrl
+      ? "public"
+      : stanWasmServerUrl === localUrl
+        ? "local"
+        : "custom";
```
WardBrian, line 32: We repeat this snippet of code in two places, might be worth pulling into a function

gui/tsconfig.json
```diff
@@ -37,15 +37,12 @@
       "@SpCore/gists/*": ["app/gists/*"],
       /* We are playing a bit fast and loose with the distinction */
       "@SpPages/*": ["app/pages/HomePage/*"],
-      "@SpStanc/*": [
-        "app/Stanc/*",
-        "app/CompilationServerConnectionControl/*",
-        "app/compileStanProgram/*"
-      ],
+      "@SpStanc/*": ["app/Stanc/*", "app/CompilationServerConnectionControl/*"],
```
WardBrian, line 40: I could debate whether or not CompilationServerConnectionControl should also be grouped with this context, but I'm fine either way in this PR

WardBrian (approved):
One suggestion but I'm happy to merge

gui/tsconfig.json
magland: This is also in main branch. Let's address separately.

gui/src/app/CompilationServerConnectionControl/ConfigureCompilationServerDialog.tsx
magland: Done

Issue #186, opened 2024-07-31
Consistent color scheme


WardBrian:
There are a few hard-coded colors, but the best thing to do is probably customize the MUI theme: https://mui.com/material-ui/customization/color/

_Originally posted by @magland in https://github.com/flatironinstitute/stan-playground/issues/159#issuecomment-2261279812_

WardBrian:
It might make sense to do this at the same time as we are deploying the logo (#9) with whatever colors we choose for that

WardBrian:
This tool: https://zenoo.github.io/mui-theme-creator/ seems useful for the MUI part of this

PR #193, opened 2024-08-01
Add help button to stan editor - open user guide

### LINKED ISSUE #192: links to Stan docs in app?
Do we want to have a link to Stan documentation somewhere in the app to help the user in writing the model?

magland:
Close #192


jsoules (approved):
Looks good to me. Only thing to consider is making the help button the same blue color as the other icons in the toolbar--wasn't sure if the difference was intentional (to make it stand out) or not.

magland:
> Looks good to me. Only thing to consider is making the help button the same blue color as the other icons in the toolbar--wasn't sure if the difference was intentional (to make it stand out) or not.

I was just being consistent with the help buttons on the data script editors. If others have a preference it doesn't matter to me.

WardBrian:
I think we can consider colors as part of #186, for now I agree matching the others is good

PR #195, opened 2024-08-01
Compile model button js

jsoules:
This refactors the heaviest branch of the RunPanel content pane into a separate file, and simplifies the content selection logic for the different branches.


magland:
Tested and seems to be working, and code is certainly cleaner

PR #196, opened 2024-08-01
Move validSyntax logic inside useStanc

WardBrian:
My suggestion from https://github.com/flatironinstitute/stan-playground/pull/191#discussion_r1700193692


jsoules (approved):
Seems reasonable to me.

magland:
Looks good

PR #203, opened 2024-08-05
Improve formatting of draws table

WardBrian:
This rounds to 6 digits and makes it so that resizing the tab doesn't lead to very skinny columns that wrap.

Before:
image

After:
image

### DIFF
--- a/gui/src/app/SamplerOutputView/SamplerOutputView.tsx
+++ b/gui/src/app/SamplerOutputView/SamplerOutputView.tsx
@@ -174,7 +174,7 @@ const DrawsView: FunctionComponent<DrawsViewProps> = ({
               <td>{drawChainIds[i]}</td>
               <td>{drawNumbers[i]}</td>
               {draws.map((draw, j) => (
-                <td key={j}>{draw[i]}</td>
+                <td key={j}>{draw[i].toPrecision(6)}</td>
               ))}
             </tr>
           ))}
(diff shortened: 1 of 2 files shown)


WardBrian:
Not sure if there is a good way to avoid the `treedepth` column (which is integer-valued) from getting all the .00000s -- @jsoules?

A very-JS suggestion I've seen online is "divide by 1 afterwards", which 1. makes TSC unhappy, but 2. does seem to work, hilariously

jsoules:
> Not sure if there is a good way to avoid the `treedepth` column (which is integer-valued) from getting all the .00000s -- @jsoules?
> 
> A very-JS suggestion I've seen online is "divide by 1 afterwards", which 1. makes TSC unhappy, but 2. does seem to work, hilariously

It's JavaScript, of course it makes no sense and yet works!

So a couple options here, but they're going to involve writing a helper function. Question is just what that helper function would look like. One solution would be to check if the number is an integer (or within tolerance of an integer) and if so don't do the `toPrecision` call. You could also do some postprocessing on the string to trim right-hand-side zeroes, up to and including a decimal point if any; could result in ragged numbers depending on the situation...

magland:
How about detecting whether it's an integer column, and then using a different format function. It will give approximately the same result as having the conditional inside the format function... but I think it's more correct in the rare case where you have a mix of integers and non-integers in the same column.

WardBrian:
image

gui/src/app/SamplerOutputView/SamplerOutputView.tsx
```diff
@@ -108,7 +108,9 @@ const DrawsView: FunctionComponent<DrawsViewProps> = ({
   >(300);
   const draws2 = useMemo(() => {
```
jsoules, line 109: As long as you're changing things, might as well make this more semantic--`formattedDraws` or the like?

jsoules (approved):
This seems good to me.

magland:
Looks good to me

Issue #201, opened 2024-08-05
If a model has too many parameters, opening the "Histograms" tab can freeze the UI


WardBrian:
Possibly worth truncating at some value (say ~50), and adding a "show more" button to the bottom, similar to the draws table.

magland:
I can work on this... but wondering if you could provide a project link for this case if not too much trouble?

WardBrian:
This is where I observed it myself: [inhalers](https://stan-playground.flatironinstitute.org/?title=Bugs%20Examples%20-%20Vol1%20-%20Inhalers%20-%20Inhalers&stan=https://raw.githubusercontent.com/stan-dev/example-models/master/bugs_examples/vol1/inhalers/inhalers.stan&data=https://raw.githubusercontent.com/stan-dev/example-models/master/bugs_examples/vol1/inhalers/inhalers.data.json)

the model takes some time to sample with the defaults, but if you set it to 10 warmup/10 sampling you can still notice the freezing on the tab

PR #204, opened 2024-08-05
Plot histograms only when visible (ReactVisibilitySensor)

magland:
A possible solution to #201 that doesn't involve a "show more" button: only renders the plot if the element is visible. Uses package react-visibility-sensor.

One note, I had to hard-code a height=200px div in the case of not visible in order for this to work. I'm sure there's a better workaround.

### DIFF
--- a/gui/package.json
+++ b/gui/package.json
@@ -38,6 +38,7 @@
     "react-plotly.js": "^2.6.0",
     "react-router-dom": "^6.17.0",
     "react-use-measure": "^2.1.1",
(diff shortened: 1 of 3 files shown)


WardBrian:
This leads to some jittering during scrolling for me, as the top row of histograms will un-render and cause you to scroll back up briefly. 

We're using the "show more" pattern in the draws table, I think it makes sense to do the same here.

WardBrian:
#205

PR #207, opened 2024-08-13
Clean up compilation error handling

WardBrian:
This fixes an issue where the error is immediately removed from the UI, and also cleans up compileStanProgram.ts to show the more appropriate error message we attach to the http response if it is present. 

I noticed yesterday the compilation server window was raising a warning about a `div` inside `<p>`, so this fixes that as well.

The easiest way to test this PR is to edit the backend server so it returns a bad response. I choice to make `_stan_src_file_is_within_size_limit` unconditionally return False to test.

### DIFF
--- a/gui/src/app/CompilationServerConnectionControl/ConfigureCompilationServerDialog.tsx
+++ b/gui/src/app/CompilationServerConnectionControl/ConfigureCompilationServerDialog.tsx
@@ -87,15 +87,15 @@ const ConfigureCompilationServerDialog: FunctionComponent<
 
         {serverType === "custom" && (
           <div>
-            <p>
-              <TextField
-                variant="standard"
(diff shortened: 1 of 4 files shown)


magland:
I tested it in a different way because I didn't want to go to the trouble of modifying the compilation server. I added an intentional error in the frontent:

```
const { mainJsUrl } = await compileStanProgram(
      stanWasmServerUrl,
      data.stanFileContent + ' intentional error',
      onStatus,
    );
```

This worked in that it showed the compile error message in the toolbar. But could I request that you also log the error to the console? Compilation would be an unexpected occurrence, but if it does happen, it will be much more convenient to view it in the debug console since it can be long.

Other than that, it looks good to go.

jsoules (approved):

PR #209, opened 2024-08-21
Add basic color theming, ability to toggle light/dark

### LINKED ISSUE #186: Consistent color scheme

WardBrian:
This uses the orange from the logo proposal we've been discussing as the primary color and its complementary color as the secondary. 

The main contribution is the button which toggles between light/dark.

image

image

There are definitely items we will want to tweak the appearance of still, but I think this closes #186

### DIFF
--- a/gui/src/app/App.tsx
+++ b/gui/src/app/App.tsx
@@ -1,19 +1,14 @@
-import createTheme from "@mui/material/styles/createTheme";
-import { ThemeProvider } from "@mui/material/styles";
-import CssBaseline from "@mui/material/CssBaseline";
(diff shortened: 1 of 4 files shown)


jsoules (approved):
Seems nice. Good work!

magland:
I'm not convinced we should support both light and dark... but this looks good to me for now.

PR #211, opened 2024-08-27
Add missing case for loading analysis.R, exhaustiveness checks

WardBrian:
Fixes a bug where analysis.R files were not being loaded properly.

This also adds some uses of the typescript `never` type to try to ensure we won't have similar "missing case in switch" bugs again


magland:
Looks good to me

jsoules:
I haven't observed the original bug--am happy for you to merge without my independent verification though if you have confirmed it is now fixed.

The `never` trick for exhaustive switch cases is a nice one. Good addition. Might consider also adding to `stancWorker.ts` though we seem more likely to potentially add new message purposes/types to that one?

jsoules (approved):

PR #220, opened 2024-09-30
Test query loading

WardBrian:
This gets us up to 100% coverage on `ProjectQueryLoading.ts`

This did dig up a couple issues:

- we were never validating that the sampling options provided piece-by-piece were valid
- `JSON.stringify` does not preserve `undefine`s, so our deepCopy implementation subtly was broken
- There were a few cases where we never indicated a problem to the user

### DIFF
--- a/gui/test/app/Project/ProjectQueryLoading.test.ts
+++ b/gui/test/app/Project/ProjectQueryLoading.test.ts
@@ -1,10 +1,85 @@
+import { afterEach, describe, expect, test, vi } from "vitest";
 import {
+  fetchRemoteProject,
   fromQueryParams,
-  //   fetchRemoteProject,
   QueryParamKeys,
   queryStringHasParameters,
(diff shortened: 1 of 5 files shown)


gui/test/app/Project/ProjectQueryLoading.test.ts
```diff
@@ -78,4 +145,158 @@ describe("Project query parameter processing", () => {
...
+      const project = await fetchRemoteProject(queries);
+      expect(project).toEqual(initialDataModel);
+
+      expect(mockedConsoleError).toHaveBeenCalledWith(
+        "Unsupported project URI",
+        "notgist",
+      );
+    });
+
+    test("fetchRemoteProject errors on when gist retrieval fails", async () => {
```
jsoules, line 275:
```suggestion
    test("fetchRemoteProject errors when gist retrieval fails", async () => {
```

Issue #69, opened 2024-06-14
make tests for ess and rhat computation and create a separate package


magland:
See

_Originally posted by @WardBrian in https://github.com/flatironinstitute/stan-playground/issues/48#issuecomment-2168241276_

PR #221, opened 2024-09-30
Add tests for ess and R-hat based on Stan's

WardBrian:
Half of #69

Just using known values from Stan's tests


magland:
Looks good to me. It's a shame to check in those large files, but as long as they don't change often I think it's fine.

WardBrian:
Stan has been testing against these same files for 11 years, so I think we're probably fine there!

If we do touch this again, it will likely be to move it to another package entirely.

jsoules (approved):
It's a bit opaque, but seeing as it's imported directly from the Stan test suite and has been stable for over a decade, I'm okay to merge it.

PR #222, opened 2024-10-07
Fix analysis.R loading from gist, again

WardBrian:
Typo in query loading and an incomplete test let this slip through


jsoules (approved):
Looks good.

PR #225, opened 2024-10-08
Disable compile buttons if no server connection

### LINKED ISSUE #194: Disable compile button if not connected to compilation server

WardBrian:
Closes #194

- Moves the connection status inside the context
- Adds a tooltip explaining why the button is disabled

image


gui/src/app/CompileContext/CompileContextProvider.tsx
```diff
@@ -15,6 +15,34 @@ type CompileContextProviderProps = {
...
+          setIsConnected(true);
+        }
+      } catch (err) {
+        setIsConnected(false);
+      }
+    })();
+  }, [probeUrl, retryCode]);
+  return { isConnected, retryConnection };
+};
+
```
jsoules, line 45: I don't mean to open up a can of worms by rewriting working code, but it occurs to me that once this is moved into the CompileContext, it may no longer actually need to be a hook. I'm wondering if it would be more natural/clear to write it as just a function or two, with the state (`useState`) held in the CompileContextProvider directly.

gui/src/app/Compilation/Context/CompileContextProvider.tsx
```diff
@@ -8,13 +7,42 @@ import {
...
+          setIsConnected(true);
+        }
+      } catch (err) {
+        setIsConnected(false);
+      }
+    })();
+  }, [probeUrl, retryCode]);
+  return { isConnected, retryConnection };
+};
+
```
jsoules, line 45: Don't mean to open up a can of worms here by suggesting a rewrite of working code, but I'm wondering--now that this is moved into the CompileContext--whether this still needs to be a hook, or if it'd be more natural/clear to write it as just a couple functions (with the state/`useState` held direct on the CompileContextProvider, perhaps).

WardBrian: I kinda liked the encapsulation, but it does seem a bit odd to have a custom hook which is only used in one place, with that one place being it's own hook-like-thing.

Happy to go either way

jsoules: I don't have strong feelings about it, and it's working as is, so it's probably not worth any mental energy to re-envision it.

You can just ship it :smile_cat:

PR #230, opened 2024-10-29
Prevent running of model with non-parseable data.json.

### LINKED ISSUE #226: Sampling with an empty Data window causes TypeError
To reproduce:

- Open Stan-Playground (e.g. https://stan-playground.flatironinstitute.org/)
- Load linear regression example from the sidebar
- Delete the contents of the "data" pane and save
- Attempt to run sampling; observe error

Unclear whether this is happening in the front-end or if we're being more tolerant of malformed data than the tinystan back-end is. I do notice that the issue occurs even if the data window string is not completely empty (e.g. contains only a space) and so is not strictly lengthless/null.

Also note, a different error is generated if the data consists only of an empty object (`{}`).

jsoules:
Fix #226.

This PR makes the `handleRun` callback (defined in `RunPanel.tsx` and used as a parameter to the `CompiledRunPanel` component) nullable. The callback will be `undefined` if the `data.json` does not parse successfully.

In turn, in the `CompiledRunPanel` component, the `Run Sampling` button will be disabled if the `handleRun` parameter is `undefined`. This prevents a downstream error in model execution caused by the failed parse.

I've added a tooltip to the "Run sampling" button that explains to the user why it is disabled in the data-did-not-parse case, and describes the function in the success case. (I also covered the cases where the model is loading or sampling is already in progress.

### DIFF
--- a/gui/src/app/RunPanel/CompiledRunPanel.tsx
+++ b/gui/src/app/RunPanel/CompiledRunPanel.tsx
@@ -2,8 +2,9 @@ import Box from "@mui/material/Box";
 import LinearProgress, {
   LinearProgressProps,
 } from "@mui/material/LinearProgress";
+import Tooltip from "@mui/material/Tooltip";
 import Typography from "@mui/material/Typography";
-import { FunctionComponent } from "react";
+import { FunctionComponent, useMemo } from "react";
 
 import Button from "@mui/material/Button";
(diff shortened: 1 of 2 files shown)


jsoules:
On further conversation with @WardBrian, it's desirable to allow an empty string in data.json. See updated PR.

PR #235, opened 2024-10-30
Make the default tab 2-spaces

WardBrian:
This aligns with what the auto-formatter's behavior is. A power user can still change it by pressing `F1` to pull up the monaco command pallet.


jsoules (approved):
Implementation is trivially fine; but the "monaco power user" solution to reset it is not at all obvious.

We should probably document it somewhere. This could be a mouseover (though I don't know of what) or the really obvious place would be in the newly expanded documentation we've added; probably in the currently-non-extant "instructions" page.

Recommendation: ship this, but also make an issue to document it later.

WardBrian:
Yes -- we should also describe that e.g. Ctrl+Enter will run the scripts

PR #237, opened 2024-10-31
Label case where data is from a prior version of generator script.

jsoules:
@WardBrian's earlier PR is a great idea, but I realized there was a risk of miscommunication with the user in the case where the data in `data.json` came from a data generation script that had subsequently been modified.

This PR expands on that one by adding data source states & logic to label this case.


WardBrian:
I do think it is good to update when the scripts themselves change, but I do wonder if it's worth actually saying "a prior version of", or if just removing the label in this case would be just as good?

jsoules:
Well, being explicit helps set the expectations a little more precisely. But I don't have a super strong feeling about it.

gui/src/app/Project/ProjectReducer.ts
```diff
@@ -71,8 +71,12 @@ const ProjectReducer = (s: ProjectDataModel, a: ProjectReducerAction) => {
     }
     case "commitFile": {
       const newState = { ...s };
-      if (a.filename === ProjectKnownFiles.DATAFILE) {
-        newState.meta = { ...s.meta, dataSource: undefined };
+      const newDataSource = confirmDataSourceForCommit(
+        s.meta.dataSource,
+        a.filename,
+      );
+      if (newDataSource !== s.meta.dataSource) {
```
WardBrian, line 78: If they _are_ equal, then the next line is a no-op right? So we could just not check?

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -124,30 +124,47 @@ const LeftView: FunctionComponent<LeftViewProps> = () => {
...
+        return "data.R";
+      }
+      case DataSource.GENERATED_BY_STALE_PYTHON: {
+        return "a prior version of data.py.";
+      }
+      case DataSource.GENERATED_BY_STALE_R: {
+        return "a prior version of data.R.";
+      }
+      default:
+        return defaultSourceDesc;
```
WardBrian, line 151: Could this be one of our "unreachable" calls?

gui/src/app/pages/HomePage/HomePage.tsx
```diff
@@ -124,30 +124,47 @@ const LeftView: FunctionComponent<LeftViewProps> = () => {
...
     } else {
       return [
         {
           type: "text",
-          label:
-            "Data is generated by data." +
-            (data.meta.dataSource === DataSource.GENERATED_BY_PYTHON
-              ? "py"
-              : "R"),
+          label: `Data is generated by ${sourceDesc}`,
```
WardBrian, line 162: Thoughts on also changing the color when stale? To something more warning-ish?

WardBrian (commented):
Slept on it and I think it makes sense to make the distinction. A couple questions about the "how":

gui/src/app/Project/ProjectReducer.ts
jsoules: Not exactly. The spread operator is creating a (shallow) copy of the object. So we'd be setting `newState.meta` to a new object with the exact same values as the original one.

In this case I don't think that matters at all, but I try to avoid it on principle because memoization checks depend on referential integrity rather than practical equivalence, so I tend to be suspicious of it.

gui/src/app/pages/HomePage/HomePage.tsx
jsoules: It is reachable though--this branch handles the case where `dataSource` is `undefined`.

I briefly considered wrapping this up further to remove the branch in the `dataMessage` memoization right below, but it felt like that was maybe a bit too cute.

gui/src/app/pages/HomePage/HomePage.tsx
jsoules: I could get behind that! Just didn't want to be too ambitious :smile:

gui/src/app/pages/HomePage/HomePage.tsx
WardBrian: Ah, fair point. I think it's the defaultSourceDesc string variable that bothers me -- would it make sense to make this a `string | undefined` rather than the empty string being special?

jsoules: I suppose making it possibly `undefined` means we always have to keep the explicit branch check in constructing the message, vs. we could set it up so that the `sourceDesc` always has the right string and then there's no need to branch in l156-157 below.

But I have very little emotional investment in this choice :)

jsoules: I was assuming that an empty toolbar would not be visible to the user.

Apart from that, this seems like a pretty minor distinction and I'm happy to do it whichever way you prefer.

WardBrian (approved):
Thanks!

PR #242, opened 2024-11-04
Make user confirm on first upload

### LINKED ISSUE #229: Indicate to users the privacy implications of using the public server
I think this could be something at the bottom of the screen, or possibly just a pop up the first time you click "compile" when connected to the public server that makes sure users understand that their code is being uploaded, but data etc are not

WardBrian:
Closes #229

### DIFF
--- a/gui/src/app/Compilation/Context/CompileContextProvider.tsx
+++ b/gui/src/app/Compilation/Context/CompileContextProvider.tsx
@@ -46,6 +46,29 @@ const useIsConnected = (stanWasmServerUrl: string) => {
 const initialStanWasmServerUrl =
   localStorage.getItem("stanWasmServerUrl") || publicCompilationServerUrl;
 
+const showOneTimeMessage = (url: string) => {
+  if (url !== publicCompilationServerUrl) {
+    // if the user opted in to a custom URL, we assume they are good with it...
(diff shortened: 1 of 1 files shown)


gui/src/app/Compilation/Context/CompileContextProvider.tsx
```diff
@@ -46,6 +46,29 @@ const useIsConnected = (stanWasmServerUrl: string) => {
...
+
+  const alreadyConfirmed = "compileModelUploadMessage";
+  if (localStorage.getItem(alreadyConfirmed) === "true") {
+    return true;
+  }
+  if (
+    window.confirm(
+      "This will upload the main.stan file to the server " +
+        "for compilation. All other files remain local.\n" +
+        "Do you want to continue? (This message will not be shown again.)",
```
jsoules, line 63:
```suggestion
        "Do you want to continue? (If you accept, this message will not be shown again.)",
```

jsoules: (Since we will keep asking if they say no but keep trying to click the button.)

jsoules (approved):
This seems fine, and works for me when I check out the local branch.

PR #243, opened 2024-11-04
Don't cache data contents in script state

WardBrian:
In 8de418f617e51ddd6df01aadf24fc49c26731848, I made it so the data.py/R scripts wouldn't update the text if the data they just generated was exactly the same. This was a nice-enough quality of life feature, but it caused a lot of unnecessary re-renders, which I "fixed" in 805ca1b67dbb4d88c2e34013ea8ee4b5b648b379.

However, this has several other bugs if the output of your script is deterministic. The easiest to see is:
1. Run a deterministic data.R
2. Edit data.json, save
3. Re-run data.R, observe that it claims the data is unchanged. 

I considered several refactors to save this feature, but all seemed too complicated to justify their use, so I'm proposing we just ditch it and always update the data after a successful script run.


jsoules (approved):
The "you already had that deterministic output around, friend" notice is a small sacrifice for code simplicity. Looks good, feel free to push

PR #244, opened 2024-11-05
Improve errors from pyodide by using filename

WardBrian:
This also uses the async call function, as this lets users call micropip

### DIFF
--- a/gui/src/app/Scripting/Analysis/AnalysisPyWindow.tsx
+++ b/gui/src/app/Scripting/Analysis/AnalysisPyWindow.tsx
@@ -59,6 +59,7 @@ const AnalysisPyWindow: FunctionComponent<AnalysisWindowProps> = ({
           loadsDraws: true,
           showsPlots: true,
           producesData: false,
+          filenameForErrors: FileNames.ANALYSISPYFILE,
         },
         files,
       });
--- a/gui/src/app/Scripting/DataGeneration/DataPyWindow.tsx
+++ b/gui/src/app/Scripting/DataGeneration/DataPyWindow.tsx
@@ -45,6 +45,7 @@ const DataPyWindow: FunctionComponent<Props> = () => {
           loadsDraws: false,
           showsPlots: false,
           producesData: true,
+          filenameForErrors: FileNames.DATAPYFILE,
         },
       });
     },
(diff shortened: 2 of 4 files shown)


jsoules (commented):
Seems reasonable; are there particular behavior changes we should verify?

WardBrian:
Two changes:

1. Tracebacks now mention the name of the file, rather than the `<exec>` string:
  Before:
    ```
   File "<exec>", line 2
    ,
    ^
   SyntaxError: invalid syntax
   ```
   Now:
   ```
    File "data.py", line 2
    ,
    ^
   SyntaxError: invalid syntax
   ```

2. You can use top-level `await`s, which lets you use micropip
   ```python
   import micropip
   await micropip.install("scikit-learn")
   ```

jsoules:
Tried the following code in `data.py`:

```python
import asyncio

async def foo():
  print("Hello")
  await asyncio.sleep(1)
  print("...world!")


await foo()
```

Before the change:
```
PythonError: sys:1: RuntimeWarning: coroutine 'foo' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
Traceback (most recent call last):
  File "/lib/python312.zip/_pyodide/_base.py", line 501, in eval_code
    .compile()
     ^^^^^^^^^
  File "/lib/python312.zip/_pyodide/_base.py", line 280, in compile
    self._gen.send(self.ast)
  File "/lib/python312.zip/_pyodide/_base.py", line 163, in _parse_and_compile_gen
    return compile(mod, filename, mode, flags=flags)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<exec>", line 10
SyntaxError: 'await' outside function
```
Note error is listed as originating in `File "<exec>"`.

After the change:
```
Hello

...world!

PythonError: Traceback (most recent call last):
  File "/lib/python312.zip/_pyodide/_base.py", line 574, in eval_code_async
    await CodeRunner(
  File "/lib/python312.zip/_pyodide/_base.py", line 396, in run_async
    await coroutine
  File "data.py", line 14, in <module>
    raise ValueError("data is not defined")
ValueError: data is not defined
```

Still an error, but only because I didn't actually generate any data for data.json. Source of the error is `File: "data.py"`.

I think this is a good change and checks out--let's ship it.

jsoules (approved):

PR #245, opened 2024-11-07
Add the option to download 'takeout scripts'

### LINKED ISSUE #179: Optional shim code when downloading a project with scripts
I think it would be nice if, when saving a file, you got an extra check box that allowed you to add a pre-written `main.py` or `main.R` to the download. 

This file would contain cmdstanpy/cmdstanr code to build the model, run it with your sampling options, and then produce objects compatible with the ones we provide in analysis.py/R

WardBrian:
Closes #179 

Currently has a working Python script, needs R and probably a better UI


gui/src/app/Scripting/Runtime/load_args.py
```diff
@@ -0,0 +1,27 @@
...
+    if k == "init_radius":
+        return "inits"
+    if k == "num_warmup":
+        return "iter_warmup"
+    if k == "num_samples":
+        return "iter_sampling"
+    if k == "num_chains":
+        return "chains"
+    if k == "seed":
+        return "seed"
```
jsoules, line 16: I feel like I want this to be a dictionary, but I think I actually just want Python to be Typescript.

gui/src/app/Scripting/Runtime/load_args.py
```diff
@@ -0,0 +1,27 @@
...
+
+    raise ValueError(f"Unknown sampling option: {k}")
+
+
+if os.path.isfile(os.path.join(HERE, "sampling_opts.json")):
```
jsoules, line 21: Where is HERE defined?

jsoules: Never mind, I see, this is meant to be part of a script that's getting assembled later.

gui/src/app/Scripting/Runtime/makePyRuntime.ts
```diff
@@ -0,0 +1,65 @@
...
+import spLoadConfig from "./load_args.py?raw";
+import spCmdStan from "./cmdstan.py?raw";
+import spRunSampling from "./sample.py?raw";
+import spDrawsScript from "../pyodide/sp_load_draws.py?raw";
+import spRunAnalysis from "./run_analysis.py?raw";
+
+const indent = (s: string) => {
+  return s
+    .trim()
+    .split("\n")
```
jsoules, line 14: It wouldn't be valid for input data to include data fields with embedded newlines, right?

gui/src/app/Scripting/Runtime/makePyRuntime.ts
```diff
@@ -0,0 +1,65 @@
...
+    .map((x) => "    " + x)
+    .join("\n");
+};
+
+const makePyRuntimeScript = (project: ProjectDataModel) => {
```
jsoules, line 19: This kind of code is always going to be pretty ugly to write, but when you're ready to take this out of draft we can chat about whether there's anything to be done with it.

jsoules (commented):
Not sure how much commentary you're looking for at this point, but I saw the draft and wanted to make sure you feel seen and heard

gui/src/app/Scripting/Runtime/makePyRuntime.ts
WardBrian: We're only calling this on text we completely control at the moment, but this is worth worrying about if the usage of this function changes

gui/src/app/Scripting/Runtime/load_args.py
jsoules: Particularly through Github :)

WardBrian:
@jsoules this is ready for review (with no particular urgency)

gui/src/app/core/Scripting/Takeout/makePyRuntime.ts
```diff
@@ -0,0 +1,60 @@
...
+  const hasDataPy = project.dataPyFileContent.length > 0;
+  const hasAnalysisPy = project.analysisPyFileContent.length > 0;
+
+  let script = `TITLE=${JSON.stringify(project.meta.title)}\n` + spPreamble;
+
+  // arguments
+  script += `argparser.add_argument("--install-cmdstan", action="store_true", help="Install cmdstan if it is missing")\n`;
+  if (hasDataJson && hasDataPy) {
+    script += `argparser.add_argument("--ignore-saved-data", action="store_true", help="Ignore saved data.json files")\n`;
+  }
```
jsoules, line 23: I think for simplicity's sake it's fine to have this parameter exist even if it can't do anything effective?

gui/test/app/Scripting/makePyRuntime.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      dataPyFileContent: "",
+    };
+    const runPy = makePyRuntimeScript(noData);
+
+    // we expect the same output minus the data loading part
+    const lines = full.split("\n");
+    const dataless =
+      lines.slice(0, 11).join("\n") +
+      "\n" +
+      lines[12] +
```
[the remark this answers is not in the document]
jsoules: Or possibly consider either breaking the semantic chunks out into separate variables here, or just importing them from the same raw source as used in actually assembling it?

I'm just thinking of ways to minimize any potential maintenance burden on these tests.

gui/src/app/core/Scripting/Takeout/makeRRuntime.ts
```diff
@@ -0,0 +1,54 @@
+import { ProjectDataModel } from "@SpCore/Project/ProjectDataModel";
```
jsoules, line 1: I assume it's not really workable to try to combine this file with the one that handles the Python runtime?

WardBrian: It turns out it wasn't _too_ bad. Let me know what you think

gui/test/app/Scripting/makePyRuntime.test.ts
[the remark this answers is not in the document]
WardBrian: These tests are certainly finicky, but I wanted to avoid essentially re-coding the logic used to generate the script in the test itself...

I think a reasonable alternative is to avoid this interpolation at all and just store the complete output for each example in the test file, as a real 'gold standard' test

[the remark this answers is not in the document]
jsoules: Yeah, I mean that's fair. It's really down to balancing the maintenance burden of having to update the golden test, vs whatever risk there is in the same person reimplementing the same logic.

I think, in light of the fact we really don't expect this code to change much, it's fine to just do as you've done here.

gui/src/app/core/Scripting/Takeout/makeRRuntime.ts
jsoules: This actually looks pretty nice.

I think if I were writing it I'd consider breaking the three commented sections ("data", "run sampling", and the other one) into their own functions, and then leaning more into interpolation rather than string-concatenation for each of the individual sections. But that's a mere stylistic difference; what you have is perfectly fine also.

WardBrian: multi-line strings and interpolation tend to look weird in this context where the output is indentation-sensitive. You end up with weird dedents in the middle of blocks that I don't love

jsoules: Fair

PR #249, opened 2024-12-04
Normalize newlines before passing to scripts

WardBrian:
Reported here: https://discourse.mc-stan.org/t/stan-playground-issues-with-pasting-scripts/37498

WebR complains about invalid tokens if given windows CRLF line endings. The easiest way to test this on non-windows is save a file in your editor using these endings, and then use the load project window to upload it (copy-pasting seems to remove them, on my Ubuntu machine at least)

I looked through the Monaco docs to see if there was a way we could ask it to do this normalization for us, but it seems there isn't, so I just put it right before we actually shell out to webr or pyodide (pyodide doesn't seem to mind either way)


magland:
Looks good to me.

jsoules:
Addresses #250

PR #252, opened 2024-12-04
Use Split component to allow sampler output view to take up more vertical space

### LINKED ISSUE #251: Provide slider to change relative sizes of model and data-gen/sampling windows
From Stan Forums [https://discourse.mc-stan.org/t/stan-playground-issues-with-pasting-scripts/37498](https://discourse.mc-stan.org/t/stan-playground-issues-with-pasting-scripts/37498):

> Also, a great improvement to the interface would be the ability to collapse the sampling/data generation window. You can currently shrink the stan/data window with the vertical bar, but there is no horizontal bar to shrink the sampling/data generation to give the analysis tab more real estate. Becomes a bigger issue when working on a small laptop screen.

WardBrian:
The default appearance is very similar to the current site, but you can now utilize basically the entire right panel for analysis if you want to
image

Closes #251

### DIFF
--- a/gui/src/app/pages/HomePage/SamplingWindow/SamplingWindow.tsx
+++ b/gui/src/app/pages/HomePage/SamplingWindow/SamplingWindow.tsx
@@ -1,7 +1,7 @@
 import { FunctionComponent, useCallback, useContext } from "react";
 
+import { Split } from "@geoffcox/react-splitter";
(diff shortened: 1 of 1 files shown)


jsoules (approved):
Code changes are straightforward; loaded it up to test and everything performs as expected. I think this is good to merge.

PR #253, opened 2024-12-10
Update Stan to 2.36.0

WardBrian:
Stan 2.36 released [this morning](https://discourse.mc-stan.org/t/release-of-cmdstan-2-36-0/37569).

This PR:

- Updates the backend to use a newer commit of tinystan which includes Stan 2.36
- Updates our copy of stanc.js

Also updates the docker image to use a newer emscripten and the new repo link for TBB.

To verify, you can use a new function like

```stan
generated quantities {
  real rng = beta_neg_binomial_rng(0.2,0.4,0.3);
}
```

PR #256, opened 2024-12-18
Add 'restart' endpoint to server for deploy automation

WardBrian:
This was discussed during #253.

This lets us ping `/restart` with a specific secret to trigger the public server to restart. If the token is not set by the user, this endpoint never does anything.

To actually make use of this, we need to manually re-deploy after a merge, picking some secret to put into both the k8 config and this repo's `production` environment secrets

### DIFF
--- a/.github/workflows/docker_deploy.yml
+++ b/.github/workflows/docker_deploy.yml
@@ -37,3 +37,7 @@ jobs:
           TAG=${INPUT_TAG:-"latest"}
           docker build -t ghcr.io/flatironinstitute/stan-wasm-server:$TAG .
           docker push ghcr.io/flatironinstitute/stan-wasm-server:$TAG
+
+      - Name: Ping public server to restart
+        run: |
+          curl -X POST https://stan-wasm.flatironinstitute.org/restart -H "Authorization: Bearer ${{ secrets.PUBLIC_SERVER_RESTART_TOKEN }}"
--- a/backend/stan-wasm-server/run.sh
+++ b/backend/stan-wasm-server/run.sh
@@ -2,4 +2,4 @@
 
 set -ex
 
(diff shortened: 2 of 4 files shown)


WardBrian:
To test:

```
cd backend/
$ docker build -t stan-playground . 
$ docker run -p 8083:8080 -e SWS_RESTART_TOKEN=4321 -it stan-playground 
```

Then, in another window
```
$ curl -X POST http://localhost:8083/restart -H "Authorization: Bearer 4321"
```

magland:
Cool! But did you test restart with uvicorn?

WardBrian:
> But did you test restart with uvicorn?

What do you mean?

At the end of the curl command I posted in my comment above, the docker image exited completely, so as long as the orchestrator restarts it we should get the behavior we want

magland:
Oh okay, uvicorn is running inside the container. Got it.

WardBrian:
Yes, inside the image uvicorn spawns several workers. When a worker gets an authenticated `restart` POST, it gets the parent process id (which will be uvicorn) and sends it an interrupt. If you look at the logs while this happens, uvicorn interprets this exactly like a Ctrl+C in the terminal, and gracefully shuts down after waiting for any active jobs to finish or the timeout elapses.

Once uvicorn exits, the docker container does too

magland:
> Yes, inside the image uvicorn spawns several workers. When a worker gets an authenticated `restart` POST, it gets the parent process id (which will be uvicorn) and sends it an interrupt. If you look at the logs while this happens, uvicorn interprets this exactly like a Ctrl+C in the terminal, and gracefully shuts down after waiting for any active jobs to finish or the timeout elapses.
> 
> Once uvicorn exits, the docker container does too

Perfect. Seems like this is the right solution.

PR #257, opened 2025-01-02
Update backend to remove job folders, simplify interface

### LINKED ISSUE #187: Clean up job folders after usage is complete
We want to keep the compiled targets for caching, but the actual job folder is still a one-and-done

WardBrian:
Closes #187 

This makes two changes:
1. Rather than serve from the job directory, the compilation step returns the hashed model id, so we can serve from there directly. Now, we can delete the job directory when it's done being used!
2. Observing that we always create, upload one file, compile, and delete, without any job re-use, we can reasonably collapse all the job endpoints into one. This is mostly useful because it removes a lot of error paths in the code: the user can never provide a non-existent job ID, for example, because they're never given one at all.

This also fixes an issue with the download routes that they were not correctly responding to HEAD requests


WardBrian:
> Given the simplification to job handling, I think we can get rid of the local cache

I agree, and I hardly ever turn down a reason to remove code...

PR #261, opened 2025-01-23
Expand settings window, add pedantic mode support

### LINKED ISSUE #227: Option for pedantic mode warnings
The Stan compiler can provide additional, "pedantic" warnings which will do things like warn against having multiple priors on the same variable. 

These can be noisy (and even downright incorrect, since some are heuristic), so it would make the most sense to have this be an opt-in of some kind. Maybe there's an options pane for the text editor, or maybe we consider making the server connection window a more general "settings" window?

WardBrian:
This closes #227.

## User-facing changes

I expanded the compilation server selection window to be a general settings window with two tabs.

Tab One: Compilation settings (basically same as before)
image

Tab two: Personalization (includes dark mode, pedantic mode, and a relocated version of the "clear all" button):

image

Pedantic mode basically triggers a lot of additional warnings, e,g:

image

## Internal changes

This ended up touching a lot of the code, mainly to avoid the continued expansion of the number of Contexts/Providers there are. 

It introduces a UserSettings context which contains the theme, pedantic mode setting, and the compilation server that is selected. This subsumes the previous ToggleableThemeContext and part of the compilation context, which does still exist in a smaller form to provide the actual connection to the server.


gui/src/app/Settings/UserSettingsProvider.tsx
```diff
@@ -0,0 +1,155 @@
...
+const colorPalette = {
+  primary: {
+    main: "#E16D44",
+  },
+  secondary: {
+    main: "#44B8E1",
+  },
+} as const;
+
+const defaultSettings = JSON.parse(
```
jsoules, line 32: It occurs to me that it'd be clearer to call this "savedSettings" or "storedSettings" or something--as they aren't defaults, they're what we've saved from the user's explicitly stated preferences.

gui/src/app/Settings/UserSettingsProvider.tsx
```diff
@@ -0,0 +1,155 @@
...
+  }, []);
+
+  // ------------------- Pedantic -------------------
+  const defaultPedantic = useMemo(() => {
+    if (defaultSettings?.pedantic) {
+      return true;
+    }
+
+    return false;
+  }, []);
```
jsoules, line 97: More stray thoughts: we do this kind of query-defaultSettings-with-a-possible-further-default for 3-4 different keys; might it be worth popping into a function/custom hook?

gui/src/app/Settings/SettingsWindow.tsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export type SettingsTab = "compilation" | "personalization";
+
+const SettingsWindow: FunctionComponent<SettingsWindowProps> = () => {
+  const {
+    settingsWindow: { isOpen, closeSettings, settingsTab },
+  } = useContext(UserSettingsContext);
+
+  let override = useMemo(() => {
```
jsoules, line 19: Not `const`?

gui/src/app/Settings/PersonalSettingsDialogue.tsx
```diff
@@ -0,0 +1,94 @@
...
+          }}
+        >
+          Clear all data
+        </Button>
+        <FormHelperText>
+          Reset all settings to their default values and clear the current
+          model. <br />
+          Make sure to download or export your work first!
+        </FormHelperText>
+      </FormControl>
```
jsoules, line 89: Do we still want to provide this? I noted in prior versions we'd kind of talked about it as being only for development.

gui/src/app/Settings/PersonalSettingsDialogue.tsx
```diff
@@ -0,0 +1,94 @@
...
+          <a
+            href="https://mc-stan.org/docs/stan-users-guide/using-stanc.html#pedantic-mode"
+            target="_blank"
+            rel="noreferrer"
+          >
+            &quot;Pedantic mode&quot;
+          </a>{" "}
+          is an option in the Stan compiler to enable additional warnings. These
+          can be instructive, but they can also lead to false positives. Be
+          prepared to take them with a grain of salt.
```
jsoules, line 59: I have confirmed that these are warnings, and thus do not block compilation. (Which would be annoying for spurious ones, even for people who are asking for this.)

jsoules: In a fun twist, it looks like one example of a spurious error shows up in the SIR model example, where line 30 complains about a function-valued argument as being potentially uninitialized :smile:

jsoules (approved):
Read through the code, loaded up the GUI and kicked the tires a little bit.

This all seems to make sense to me. I think it makes several things tidier, too.

I had a couple stray thoughts, but you can also just push as is.

gui/src/app/Settings/PersonalSettingsDialogue.tsx
WardBrian: I like having it around still

PR #262, opened 2025-01-23
Add console text output tab

### LINKED ISSUE #260: Should print statements end up somewhere other than the developer console?
Currently, these are not really visible/discoverable to the average user. 

We could consider adding a tab under "Output" that has these messages. 

It may be tricky to do these without bogging down the worker communication or triggering a lot of re-renders if we implement it naively -- perhaps the worker can store the messages and then return the sum total of them when it's done?

WardBrian:
Closes #260 

I ended up _not_ doing something live because this would dramatically slow down the sampling if a print statement was present in the model block. Now it only _slightly_ slows it down, at the cost of not being live. If a user really wants to monitor the output live, they still can use the developer console.

image

### DIFF
--- a/gui/src/app/SamplerOutputView/SamplerOutputView.tsx
+++ b/gui/src/app/SamplerOutputView/SamplerOutputView.tsx
@@ -1,4 +1,7 @@
 import { Download } from "@mui/icons-material";
+import Box from "@mui/material/Box";
+import Button from "@mui/material/Button";
+import IconButton from "@mui/material/IconButton";
 import Table from "@mui/material/Table";
 import TableBody from "@mui/material/TableBody";
 import TableCell from "@mui/material/TableCell";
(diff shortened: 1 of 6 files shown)


jsoules:
Having looked at nothing yet--I assume the massive slowdown for live updates comes from React and rerendering all the time; is it possible we could escape-hatch this by doing some set-inner-html nonsense?

Not saying it'd be a good idea, mind, but...

WardBrian:
Hm, I thought the slowdown was also on the worker thread, but you may be right. I’ll try tomorrow through the innerHTML side door

jsoules:
No idea if it would help/how much. In the screenshotted example I'm noticing a slowdown of ~2.5 seconds per run over 4-5 runs, for 4 chains, 1k warmup 1k samples, radius 2. (That about doubles the wall clock time, but it's a trivial example.)

Whereas for the SIR model example, with 100 samples, it's a negligible amount--much less than the inter-run variation without the print statement.

jsoules (commented):
This is probably good to go (I tried it out locally as well), but I pointed out a couple opportunities for refinement if you should choose to pursue them.

WardBrian:
For posterity: innerHTML trickery _was_ faster than reactivity for live streaming of output, but was still a decent factor slower than I think we'd be willing to tolerate

PR #263, opened 2025-01-24
Chores: Version bumps

WardBrian:
This bumps a few different versions, including emscripten, pyodide, and webr.

I confirmed that the functionality is unchanged by essentially just running everything in the disease transmission example, including the scripts


jsoules (approved):
I can't swear that I've tested all the features we've implemented, but running this locally, I was able to successfully compile and sample from both demo models, and use both data generation and analysis scripts in both python and R.

Reviewing the changes, everything makes sense.

I think we're good to go here.

PR #264, opened 2025-01-27
Show which files are unsaved on disabled save/load buttons

WardBrian:
image

This also cleans up a bit of dataflow where a component was receiving a redundant piece of information as an argument which it already had access to internally

### DIFF
--- a/gui/src/app/Project/ProjectDataModel.ts
+++ b/gui/src/app/Project/ProjectDataModel.ts
@@ -1,4 +1,5 @@
 import baseObjectCheck from "@SpUtil/baseObjectCheck";
+import { ProjectFileMap } from "@SpCore/FileMapping";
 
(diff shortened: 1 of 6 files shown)


jsoules (commented):
Code changes are reasonable and the behavior in the UI appears to match what is intended.

However, it seems a little weird that the "save project" button gets disabled if there are unsaved files. On reflection it makes sense; "Save Project" in that context means "persist the project to one of our sharing/reloading mechanisms," not "save the unchanged changes in the editor," and it's good to disallow an action when the results would be potentially unintuitive.

But I wonder if there's another word we can use for the "save project" button to get around the friction of a "save" button complaining about unsaved content...

WardBrian:
"Export" seems like the obvious synonym?

PR #265, opened 2025-01-29
Improve LazyPlotlyPlot lagging on resize

WardBrian:
This uses the `debounce` option to `useMeasure` to make it so the plots don't get re-drawn quite so often when resizing the output view. The UI lags significantly less now when the "histograms" output is active and the slider is dragged to widen that side of the screen.

I also tweaked the visuals of the fallback, since I spent some time staring at them during testing

### DIFF
--- a/gui/src/app/components/LazyPlotlyPlot.tsx
+++ b/gui/src/app/components/LazyPlotlyPlot.tsx
@@ -1,9 +1,9 @@
 import CircularProgress from "@mui/material/CircularProgress";
-import React, { FunctionComponent, Suspense } from "react";
+import React, { FunctionComponent, Suspense, useMemo } from "react";
+import useMeasure from "react-use-measure";
 
 import type { PlotParams } from "react-plotly.js";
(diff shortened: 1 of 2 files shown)


jsoules:
In general I think this is a good idea, and I can definitely see the improved smoothness.

I'm not 100% about moving the `div` with the `ref` to inside the `Suspense` block--doesn't this mean that you can't start measuring until Plotly loads? It might just be my eyes tricking me, but I feel like there's an extra frame of resizing/adjustment after the load that isn't there in the current pub. (Or I might just be noticing it more because of the debounce.)

Similarly, if you look at the line plots (which... I'm not sure if/how they would be affected by the changes you've made here) in the published version they're starting a little too big and then snapping back down a few pixels, while in the PR version I'm seeing them start at half width and then expand noticeably. Is something giving them a better initial value in the published version?

Regardless, the debounce is definitely the right thing to do and you should move forward with it, I just wanted to flag what I noticed.

jsoules:
(Noting that per our conversation, it would in fact represent an unacceptable increase in scope)

PR #267, opened 2025-01-30
Fix downloading monaco twice per load

WardBrian:
I noticed some slow loading during a dev environment that made me inspect the network tab and realize monaco was _not_ loading from `localhost`, but from a CDN.

After some digging, it turns out `@monaco-editor/react` uses a package called `@monaco-editor/loader` that has this behavior. I was unsuccessful in getting it to use our bundled monaco, so instead I went the other way and made it so we don't have to bundle `monaco-editor` ourselves at all.

The bundle size is much happier now, here's yarn build:

```
dist/assets/__vite-browser-external-9wXp6ZBx.js      0.03 kB
dist/index.html                                      0.69 kB │ gzip:   0.37 kB
dist/assets/StanModelWorker-DlhrIcAU.js              6.88 kB
dist/assets/pyodideWorker-C-8uEa4H.js               19.90 kB
dist/assets/stancWorker-CJUPMKJw.js              1,943.28 kB
dist/assets/index-DAWNhffP.css                       2.51 kB │ gzip:   1.05 kB
dist/assets/mui-CgtEPZ7g.js                        322.76 kB │ gzip:  98.74 kB
dist/assets/utilities-C_bYLVjP.js                  324.77 kB │ gzip:  86.53 kB
dist/assets/index-Amr1Itqm.js                      362.06 kB │ gzip: 117.54 kB
dist/assets/plotly-cartesian-DY12K7b9.js         1,402.37 kB │ gzip: 478.09 kB
```


gui/src/app/components/FileEditor/TextEditor.tsx
```diff
@@ -18,9 +7,27 @@ import {
...
+import { Editor, loader, useMonaco, type Monaco } from "@monaco-editor/react";
+import type { editor, IDisposable } from "monaco-editor";
+
 import { ToolBar, ToolbarItem } from "./ToolBar";
 
+import monacoAddStanLang from "./monacoStanLanguage";
+// loader from @monaco-editor/react handles the loading of the monaco editor
+// importantly, it downloads from a CDN, so we need to make sure we
+// only depend on types from the monaco-editor package to avoid
+// downloading twice.
```
jsoules, line 24:
```suggestion
// importantly, it downloads from a CDN, so we need to make sure our
// dependency on the monaco-editor package is limited to types only,
// to avoid downloading twice.
```

gui/src/app/components/FileEditor/TextEditor.tsx
```diff
@@ -18,9 +7,27 @@ import {
...
+
+loader.config({
+  paths: {
+    vs: "https://cdn.jsdelivr.net/npm/monaco-editor@0.52.2/min/vs",
```
jsoules, line 28: Is this going to incur a maintenance cost for monaco version upgrades? (Skimming over the documentation, it isn't immediately clear to me why we need to specify a particular delivery path--isn't this the default?)

gui/src/app/components/FileEditor/TextEditor.tsx
```diff
@@ -97,24 +104,34 @@ const TextEditor: FunctionComponent<Props> = ({
...
+    if (!editorInstance || !monacoInstance) return;
     if (!contentOnEmpty) return;
     if (text || editedText) {
       return;
     }
-    const contentWidget = createHintTextContentWidget(contentOnEmpty);
+    const widgetPosition = {
+      position: { lineNumber: 1, column: 1 },
+      preference: [monacoInstance.editor.ContentWidgetPositionPreference.EXACT],
+    };
```
jsoules, line 115: Seems like this ought to be `const`-able, but the linter doesn't like it. Probably doesn't matter anyway...

jsoules (approved):
I'd like to clarify one of the comments, as I think people may be lost who haven't discussed the problem the way you and I have recently.

I will take you at your word about specifying the CDN URL, but is it really necessary to do so?

WardBrian:
> I will take you at your word about specifying the CDN URL, but is it really necessary to do so?

Specifying the URL is necessary if you want control over the version. So far we have (unknowingly!) always been using monaco-loader's default version, [which is 0.46 ](https://github.com/suren-atoyan/monaco-loader/blob/master/src/config/index.js#L3)

WardBrian:
From what I can tell, historically, requests for them to bump their default version are met with people being told just to use the config like this. I think they want to avoid doing an npm release every time monaco does one

gui/src/app/components/FileEditor/TextEditor.tsx
WardBrian: Why is described in below comments, but yes, this does require some upkeep (though I should note pyodide has basically the same requirement in https://github.com/flatironinstitute/stan-playground/blob/main/gui/src/app/core/Scripting/pyodide/pyodideWorker.ts#L15)

gui/src/app/components/FileEditor/TextEditor.tsx
WardBrian: Not sure I follow what you mean by const-able in this context?

jsoules: I mean annotated as `as const`, since its members don't seem to be able to be changed anywhere in the code. But that conflicts with the looser typing further down.

(The larger context was me wondering if it should be memoized, before realizing this was already in a `useEffect` hook, so it doesn't matter.)

jsoules:
What must their release schedule be like, that that would be an actual problem greater than everybody by default using a version from 2023

WardBrian:
Well, we _were_ using a version from 2023 and didn't notice... ;P

PR #268, opened 2025-01-31
Clean up typing, memoization in plotly components

WardBrian:
Primary difference is the removal of `chainColorForIndex` and just passing our list of colors to Plotly as a "colorway" to get the same effect, as well as some memo cleanup


jsoules (approved):
Seems fine. I don't notice any changes in the UI.

PR #271, opened 2025-02-04
Tweak backend settings for better binary sizes

WardBrian:
Saves about 300kb per model (now only about ~1mb in size total). No impact on runtimes (passing -Os saved an extra ~100kb, but would cost ~5% in runtime, so I didn't).

I think this may also slightly improve compile times

### DIFF
--- a/backend/local.mk
+++ b/backend/local.mk
@@ -9,11 +9,13 @@ TBB_LIB=/app/oneTBB/install/lib/
 LDFLAGS_TBB ?= -Wl,-L,"$(TBB_LIB)"
 LDLIBS_TBB ?= -ltbb
 
+TINYSTAN_SERIAL=true
+
 # could also uses -fexceptions which is more compatible, but slower
 CXXFLAGS+=-fwasm-exceptions
 
 LDFLAGS+=-sMODULARIZE -sEXPORT_NAME=createModule -sEXPORT_ES6 -sENVIRONMENT=web -sINCOMING_MODULE_JS_API=print,printErr
 LDFLAGS+=-sEXIT_RUNTIME=1 -sALLOW_MEMORY_GROWTH=1
 # Functions we want. Can add more, with a prepended _, from tinystan.h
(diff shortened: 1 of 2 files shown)


backend/local.mk
```diff
@@ -9,11 +9,13 @@ TBB_LIB=/app/oneTBB/install/lib/
...
+TINYSTAN_SERIAL=true
+
 # could also uses -fexceptions which is more compatible, but slower
 CXXFLAGS+=-fwasm-exceptions
 
 LDFLAGS+=-sMODULARIZE -sEXPORT_NAME=createModule -sEXPORT_ES6 -sENVIRONMENT=web -sINCOMING_MODULE_JS_API=print,printErr
 LDFLAGS+=-sEXIT_RUNTIME=1 -sALLOW_MEMORY_GROWTH=1
 # Functions we want. Can add more, with a prepended _, from tinystan.h
-EXPORTS=_malloc,_free,_tinystan_api_version,_tinystan_create_model,_tinystan_destroy_error,_tinystan_destroy_model,_tinystan_get_error_message,_tinystan_get_error_type,_tinystan_model_num_free_params,_tinystan_model_param_names,_tinystan_sample,_tinystan_pathfinder,_tinystan_separator_char,_tinystan_stan_version
+EXPORTS=_malloc,_free,_tinystan_api_version,_tinystan_create_model,_tinystan_destroy_error,_tinystan_destroy_model,_tinystan_get_error_message,_tinystan_get_error_type,_tinystan_model_num_free_params,_tinystan_model_param_names,_tinystan_sample,_tinystan_separator_char,_tinystan_stan_version
```
jsoules, line 20: Don't we need that for Pathfinder support?

Or I guess we don't really give people a way to access it through the front-end right now anyway...

jsoules:
I assume you mean 5% sampling performance hit for the size-optimized code, and yeah I'd agree that's a bad trade-off to save like, a meme's worth of bandwidth.

backend/local.mk
WardBrian: Yeah, we're shipping a bunch of compiled code with no chance to use it right now. We can pay for it if we ever start needing it

jsoules (approved):
This all seems fine--go ahead!

PR #275, opened 2025-02-12
Add a salt to the model output hash

WardBrian:
@jsoules brought up an issue I had also seen in #274, which I think is caused by the changes #257 made which, among other things, mean the URL for a given model was stable across backend updates/restarts. 

I think this is actually a bad thing, since it can mean a compilation result is served from the user's local cache rather than the server. Even if this isn't the source of the weird loading behavior we've both seen, it certain would cause issues if we needed to change something about the returned model's API that needed a simultaneous backend and frontend update.

This PR does the simplest fix I could think of, which is the docker build now computes a hash of all of the application files, and this hash is used to salt the model output folder generation. This means in a given build, all the workers will still compute the same hashes, but changing any file in our code or in the tinystan or tbb dependencies will lead to new URLs for the same models.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -52,8 +52,13 @@ RUN cd tinystan && \
     emmake make test_models/bernoulli/bernoulli.js -j$(nproc) && \
     emstrip test_models/bernoulli/bernoulli.wasm
 
-COPY stan-wasm-server /stan-wasm-server
-WORKDIR /stan-wasm-server
+COPY stan-wasm-server /app/stan-wasm-server
+
+# compute a hash of our app to make the model cache use unique URLS
+# (to avoid browsers caching across different versions)
+RUN find /app/ -type f -print0 | sort -z | xargs -0 sha1sum | sha1sum | cut -d' ' -f1 | tee /app/stan-wasm-server/hash-salt.txt
+
+WORKDIR /app/stan-wasm-server
 
 ENV SWS_PASSCODE=1234
 ENV SWS_LOG_LEVEL=debug
--- a/backend/stan-wasm-server/src/app/logic/compilation.py
+++ b/backend/stan-wasm-server/src/app/logic/compilation.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import time
+from functools import lru_cache
 from hashlib import sha1
 from pathlib import Path
 from shutil import copy2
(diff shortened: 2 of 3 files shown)


WardBrian:
Things to verify as the reviewer:

1. Running locally, repeated compilations are still hitting the server cache (i.e., each worker is still computing the same hash for the same model)
2. After changing a file and re-building, the returned URL is different for a given model

backend/stan-wasm-server/src/app/logic/compilation.py
```diff
@@ -18,10 +19,27 @@
...
+    """
+    salt_file = Path(__file__).parent.parent.parent.parent / "hash-salt.txt"
+    if not salt_file.exists():
+        logger.warning("No hash salt file found at %s", salt_file)
+        salt = b""
+    else:
+        salt = salt_file.read_bytes().strip()
+        logger.info("Using hash salt from %s: %s", salt_file, repr(salt))
+    return salt
+
```
jsoules, line 36: Should I be seeing the side effect of this function every time I compile a new (cache miss) `main.stan`? I would've thought these would be cache hits, but maybe I need a refresher on our threading model here.

jsoules: Actually, it's not even that predictable when the function reruns and when it doesn't.

I don't think this much matters, but I like to note everything.

jsoules (approved):
Hash seems consistent between different threads when recompiling, and I see no evidence of any compilation issues. Salt changed when I changed a file in the `wasm-server` directory, and reverted when I re-checked out the earlier version.

I think we're probably good to go here, though there is some sort of squirrelly unpredictability with when the hash-salt-lookup function gets called and when it doesn't.

backend/Dockerfile
```diff
@@ -52,8 +52,13 @@ RUN cd tinystan && \
     emmake make test_models/bernoulli/bernoulli.js -j$(nproc) && \
     emstrip test_models/bernoulli/bernoulli.wasm
 
-COPY stan-wasm-server /stan-wasm-server
-WORKDIR /stan-wasm-server
+COPY stan-wasm-server /app/stan-wasm-server
+
+# compute a hash of our app to make the model cache use unique URLS
+# (to avoid browsers caching across different versions)
+RUN find /app/ -type f -print0 | sort -z | xargs -0 sha1sum | sha1sum | cut -d' ' -f1 | tee /app/stan-wasm-server/hash-salt.txt
```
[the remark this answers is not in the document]
WardBrian: I suppose it doesn't matter much either way, but if no files changed we actually wouldn't mind a user's local cache being warm

backend/stan-wasm-server/src/app/logic/compilation.py
jsoules: Is there any rhyme or reason to when a new thread is spawned?

backend/Dockerfile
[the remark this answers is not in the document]
jsoules: Well, I suppose redeploying the server isn't exactly something we do so often that it's worth worrying about a spurious sort.

backend/stan-wasm-server/src/app/logic/compilation.py
WardBrian: Not sure, that would be a uvicorn implementation detail. It seems like it’s roughly round-robin

PR #276, opened 2025-02-12
Add a text box to the Load Project window for URLs

### LINKED ISSUE #223: Add text box to load from a gist to the "Load Project" window

WardBrian:
Closes #223 

This lets you load from a link someone sends you without just pasting it into the tool bar.

image


gui/src/app/windows/LoadProjectWindow/LoadProjectPanel.tsx
```diff
@@ -119,25 +135,43 @@ const LoadProjectPanel: FunctionComponent<LoadProjectProps> = ({ onClose }) => {
...
-            </li>
-          </ul>
-        </div>
-        <UploadFiles height={300} onUpload={onUpload} />
+        <FormControl margin="normal">
+          <TextField
+            variant="standard"
+            label="Project URL"
+            value={urlToLoad}
+            onChange={(e) => setUrlToLoad(e.target.value.trim())}
```
jsoules, line 148: I'm not sure having this automatically execute the redirect is the right approach. It's nice to get the immediate feedback on whether the target URL is valid, but I can imagine a situation where I actually want to edit the string manually or something.

Not the biggest deal in the world, but I think I'd prefer if the callback that does the navigation were separate, and triggered on the text element losing focus (or just "user hit enter").

WardBrian: I'd be happy to do whatever, the instantaneous nature is shared with if you drop a zip file in the box below but I guess that feels more final.

Do you know what the callback you're suggestion would be called? There are dozens of `onSomething` props available for this element and I don't know which does what you describes

jsoules: If you want something to execute when an element loses focus, the relevant callback is `onBlur()`. For catching the `enter` stroke, it'd be `onKeyPress()`.

jsoules (approved):
One could continue to iterate on this and make it even more helpful, but at the expense of more and more complication in terms of URL validation. I think we'll hit diminishing returns on that pretty quick--so we should probably put it off until someone is actually asking for it.

Feel free to do so if you feel like it, but I would also be fine publishing this as is.

PR #283, opened 2025-03-13
More user-friendly variable names on output views

WardBrian:
The parameter names directly out of Stan are formatted such that they can be a CSV header row, which means they avoid using `,`s. This means they don't always look like the most 'natural' representation to the user.

This performs the [canonical transformation](https://github.com/stan-dev/stan/blob/d4c6ae376bba1b754b7e7cdbe9269dce93958065/src/stan/io/stan_csv_reader.hpp#L16) for these names such that they look more like they would if the user was trying to access that element in Stan itself.

For example, `pred_cases.2`, the second element of the `pred_cases` array in the disease transmission example, will be shown as `pred_cases[2]` on the plots and summary table. Note that the draws table still uses the CSV-safe name, since we allow downloading it as one.

### DIFF
--- a/gui/src/app/areas/ControlArea/SamplingArea/ResultsArea/SamplerOutputArea/HistogramsPanel.tsx
+++ b/gui/src/app/areas/ControlArea/SamplingArea/ResultsArea/SamplerOutputArea/HistogramsPanel.tsx
@@ -2,6 +2,7 @@ import Button from "@mui/material/Button";
 import ResponsiveGrid from "@SpComponents/ResponsiveGrid";
 import Histogram from "./Histogram";
 import { FunctionComponent, useMemo, useState } from "react";
+import prettifyStanParamName from "@SpUtil/prettifyStanParamName";
 
 type HistogramsProps = {
   draws: number[][];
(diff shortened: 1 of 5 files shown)


gui/src/app/areas/ControlArea/SamplingArea/ResultsArea/SamplerOutputArea/HistogramsPanel.tsx
```diff
@@ -15,23 +16,36 @@ const HistogramsPanel: FunctionComponent<HistogramsProps> = ({
...
     // put the names that don't end with __ first
-    const names = paramNames.filter((name) => !name.endsWith("__"));
-    const namesWithSuffix = paramNames.filter((name) => name.endsWith("__"));
+    const names: [string, number][] = [];
+    const namesWithSuffix: [string, number][] = [];
+
+    for (const [index, name] of paramNames
+      .map(prettifyStanParamName)
+      .entries()) {
+      if (name.endsWith("__")) {
```
jsoules, line 25: Does this play okay with the prettifying? From how I read you here, you shouldn't have a case where a trailing __ gets hidden by bracketization since those are supposed to be numeric indices, but is that the case for tuples as well, those are always numeric?

And would we want to sort a dunder-suffixed tuple following this rule (e.g. `theta__.7 ` belongs with the regular parameters or not?) Or does that just not happen because the double-underscore-suffixed ones are system variables?

WardBrian: It is illegal in Stan to end a variable with `__`, so the only ones that show up are the algorithmic ones, and those are always scalars

jsoules: Perfect. In that case I don't see any issues on visual inspection. What's a good model to check it out with?

WardBrian: The Disease Transmission example has a bunch of array parameters. I don't think we have an example that uses tuples, but a simple one is something like

```stan
parameters {
  tuple(vector[10], real) foo;
}
model {
 foo.1 ~ normal(1,5);
 foo.2 ~ std_normal();
}
```

jsoules (approved):
Ran a couple tests with the SIR model and the simple tuple example linked in the discussion. I don't see anything obviously wrong here (other than highlighting some rather confusing choices about the output order & indexing, which aren't our responsibility here), so I think this is fine to move forward.

PR #291, opened 2025-04-03
Try codecov coverage and bundle analysis

WardBrian:


WardBrian:
It seems like this only really does anything after the first merge, so I'm going to merge and if we don't like it we can revert

PR #302, opened 2025-08-08
Uniformly sort diagnostics to the end of draws output

### LINKED ISSUE #298: More of a split between model parameters and diagnostics in plots
Feedback from @charlesm93 following a summer school:

> The playground is great. The only thing I would comment is that the default analysis includes too many HMC-specific details (like plots of energy). I would stick to standard plots (trace plots and histograms for model parameters), since that's all you need for a demo. That or give instructor some control over which plots appear.

We currently do re-sort the 'Histograms' tab to put things like energy at the bottom, but we do not do this for trace plots. Further more, it might make sense to gate all of these (except lp__?) behind a 'Show diagnostic variables' checkbox

WardBrian:
Closes #298. This is the simplest solution, which is just put all the diagnostic parameters at the end of the lists. We already did this for histograms, so just by moving it up a component we get it everywhere we care about.

### DIFF
--- a/gui/src/app/areas/ControlArea/SamplingArea/ResultsArea/SamplerOutputArea.tsx
+++ b/gui/src/app/areas/ControlArea/SamplingArea/ResultsArea/SamplerOutputArea.tsx
@@ -28,15 +28,29 @@ const SamplerOutputArea: FunctionComponent<{ latestRun: StanRun }> = ({
   const variables = useMemo(() => {
     const numChains = sampleConfig.num_chains;
 
-    return paramNames.map((name, index) => ({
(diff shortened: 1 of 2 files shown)


WardBrian:
> A factor we may not have considered is that for models with lots of actual parameters, this change will bury the diagnostics--for instance on the Lotka-Volterra model, you have to scroll quite a bit to get to actually see the lp__. I'm assuming this is an acceptable compromise from the user perspective, compared to burying model parameters in diagnostics.

Yeah, this is one argument in favor of a checkbox which hides/shows them at the top. I'm reluctant to introduce more modality though

PR #307, opened 2025-09-16
Proxy gist loading to get around rate limits

### LINKED ISSUE #306: Gist links now failing with 401
Clicking any link on our sidebar now yields 

```
GET
https://api.github.com/gists/93d12876923790f23d9c5cb481e8cd34
[HTTP/2 401  19ms]

GET /gists/93d12876923790f23d9c5cb481e8cd34 - 401 with id CDC4:230E2E:2258E0:795CB6:68C86705 in 78ms [utilities-DOgRh2tD.js:11:4189](https://stan-playground.flatironinstitute.org/assets/utilities-DOgRh2tD.js)
Failed to load content from gist HttpError: Requires authentication - https://docs.github.com/rest
```

I can't tell if this is just a rate limit or if something in the API changed. We may need to authenticate these requests?

WardBrian:
Closes #306, thanks @magland!

As an aside, would you mind setting up the CORS to allow 127.0.0.1 instead of/in addition to `localhost`? That's what we use in the backend server

### DIFF
--- a/gui/src/app/util/gists/doesGistExist.ts
+++ b/gui/src/app/util/gists/doesGistExist.ts
@@ -1,20 +1,20 @@
-import { Octokit } from "@octokit/rest";
-
 const doesGistExist = async (gistUri: string): Promise<boolean> => {
-  const parts = gistUri.split("/");
-  const gistId = parts[parts.length - 1];
-  if (!gistId) {
-    return false;
-  }
-  const octokit = new Octokit();
   try {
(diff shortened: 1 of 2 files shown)


magland:
> Closes #306, thanks @magland!
> 
> As an aside, would you mind setting up the CORS to allow 127.0.0.1 instead of/in addition to `localhost`? That's what we use in the backend server

Okay that's done @WardBrian 

And for reference, here's the source code:
https://github.com/magland/stan-playground-gist-access

PR #329, opened 2025-11-03
Write more digits from data.R, fix brms always being installed

WardBrian:
Two fixes:

1. data.R wasn't writing with high precision.  The easiest way to see this is with `data <- list(a = c(0.1 + 0.2))`, it was just outputting `0.3` rather than the correct floating point value. For real data, it seemed like it was capping out at ~4 decimals, which is too few
2. data.R was always installing brms on the second run, due to a conflict between `webr::shim_install` and `requireNamespace`. This is fixed.


jsoules (approved):
Code seems straightforward enough.

Verified that `data <- list(a = c(0.1 + 0.2))` (`c` in R being a function for making a vector) returns a rounded `0.3` in current production and the (correctly incorrect, per FP representation) `0.30000000000000004` in the proposed code.

Regarding the R `brms` double load issue, talked with Brian and this actually shows up if you try to run an R script that does *not* import brms. In that case, on current production, you'll see a bunch of installation happen the second time you try to run R code. This behavior is fixed in the proposed PR.

Have not attempted to check any performance characteristics.

I think we're good to merge this.

PR #330, opened 2025-11-03
Remove unnecessary waits

WardBrian:
As far as I can tell, these were not doing anything other than slowing things down


magland:
I think it's safe to remove.

PR #334, opened 2025-12-04
chore: bump emscripten docker image version

WardBrian:
As far as I can tell, none of the changes directly affect us in terms of needing to update our code, but the built wasm does get 5-10% smaller just by updating.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,5 +1,5 @@
 # Dockerfile
-FROM emscripten/emsdk:4.0.9 AS builder
+FROM emscripten/emsdk:4.0.21 AS builder
 
 # Set the working directory
 WORKDIR /app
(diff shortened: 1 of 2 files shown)


jsoules:
Hey, smaller payload is always good...

I've rerun all of the examples comparing live against a local compilation server and the current public server. No compilation or execution errors. Results seem to be generally in line with each other (though I didn't do some kind of statistical analysis, just eyeballing the parameters--but the top lines are all within less than a percent of each other with default settings.) I don't think a compilation performance test would be particularly meaningful given the different hardware involved (plus I'm getting cache hits from the public server) but performance for the local seemed acceptable.

I don't see any blockers here--feel free to merge & deploy, and we can always roll it back.

jsoules (approved):

PR #357, opened 2026-04-23
Remove brms-shimmed variables from global environment at start of run

WardBrian:
This fixes a subtle bug in data generation:

1. Run the brms data generation example
2. Manually edit the model, and remove the brms call from data generation
3. Re-run data generation. Observe that, unexpectedly, the model code is updated.

### DIFF
--- a/gui/src/app/core/Scripting/webR/webR_preamble.R
+++ b/gui/src/app/core/Scripting/webR/webR_preamble.R
@@ -2,3 +2,8 @@
 webr::shim_install()
 # enable the canvas backend for plotting
 webr::canvas()
+
+# cleanliness -- brms shim modifies persistent environment, clean that up
+if (exists(".SP_CODE")) {
+  remove(".SP_CODE", "data", envir = .GlobalEnv)
+}


jsoules:
Steps to reproduce:

- Load (for instance) the Disease Transmission model from the sidebar
- Open Data Generation -> data.R. Delete everything and save. On save, this populates the empty document with a link that will dump in a canned BRMS example.
- Click that link, and uncomment the BRMS calls to load the library and generate the data. Now run data.R.
- Note that the main.stan / model code has been replaced with something BRMS generated.
- Revert both main.stan and data.R by re-clicking the sidebar link. Note that BRMS is no longer in data.R's listing
- Re-run the (stock, BRMS-less) data.R and notice that it *still* overwrites the main.stan file, as it's invoking code that's in the header but no longer present in the visible data.R code (note that following the procedure described here, you will likely still get the data that would be generated by the visible data.R, it's the model that's been overwritten)

jsoules (commented):
Checked out this branch and followed the steps to reproduce. Confirmed that the issue no longer appears.

I have no problem with the code changes themselves, either, and running Data.R appears to still work, so I think we are good to merge this and push to prod when desired.

jsoules (approved):

PR #358, opened 2026-04-23
Use stan-language-server and MonacoLspClient for language enhancements

### LINKED ISSUE #321: Enhancement: Replace custom Stan editor integration with stan-language-server
https://github.com/tomatitito/stan-language-server is nearing a release. This would provide the diagnostics and formatting we already do, plus function name completion and hover documentation (and maybe more features in the future).

There is an existing [monaco-languageclient package](https://github.com/TypeFox/monaco-languageclient) that could be used for connecting these two, but it might require switching to a different react loader for Monaco, which could be a big change.

WardBrian:
Closes #321. Monaco 0.55 added a `monaco-lsp-client` which lets use replace our custom stanc integration with https://github.com/tomatitito/stan-language-server. 

The same features as before are supported (error/warning squigglies, auto-format), plus hover documentation and autocompletion
image

Unfortunately, monaco-lsp-client does not currently support the `workspace/configuration` method, which is needed to maintain the pedantic mode support we have in the current implementation. The majority of the code in `stanLsp.ts` is dedicated to adding our own patched support around this. Hopefully it can one day be deleted.


gui/src/app/components/FileEditor/TextEditor.tsx
```diff
@@ -159,18 +137,32 @@ const TextEditor: FunctionComponent<Props> = ({
...
-  const { theme: userTheme } = use(UserSettingsContext);
-
+  const { theme: userTheme, pedantic } = use(UserSettingsContext);
   const theme = useMemo(
     () => (userTheme === "dark" ? "vs-dark" : "vs"),
     [userTheme],
   );
 
+  // hack: monaco-lsp-client doesn't support workspace/diagnostic/refresh,
+  // so we need to trigger a re-diagnose an other way when pedantic mode changes
```
jsoules:
```suggestion
  // so we need to trigger a re-diagnose another way when pedantic mode changes
```

jsoules (approved):
As you suggested, looked over the code, without worrying too much about the details of stanLsp.ts (which I assume is mostly boilerplate from the Monaco library component).

I think this is good to go; it's nice to remove the need for a separate window, and the autocomplete is more full-featured and hovering seems natural. Unfortunate that most of the functions don't have any descriptive text beyond a link to the manual, but them's the breaks.

Not a blocker, but I notice that the hover-text seems to have a minimum width which may require resizing the editor window to see the whole thing. If there's an easy setting to reduce this, great; otherwise, we're fine to publish as is.

PR #359, opened 2026-05-26
Update Stan to 2.39

WardBrian:
Updates tinystan and stan-language-server

Here's a model that uses one of the new functions:

```stan
data {
  int<lower=0> N;
  vector[N] x;
  vector[N] y;
}
parameters {
  real alpha;
  real beta;
  real<lower=0> sigma;
  matrix[3,4] A;
  matrix[4,3] B;
}
model {
  to_vector(A) ~ std_normal();
  to_vector(B) ~ std_normal();
}

generated quantities {
   real foo = trace_dot(A,B);
}
```


jsoules:
Checked out the branch and ran a local ui (`yarn install` and `yarn dev` in `gui` directory) against a local backend (`docker build -t stan-playground . ; docker run -p 8083:8080 -e SWS_RESTART_TOKEN=4321 -it stan-playground` in `backend` directory).

Confirmed that compilation for the example passes both front-end and back-end checks with the local version (this branch), while it gets a syntax error on the live public branch. Sampling also runs successfully for the resulting compiled model.

Also ran sampling for the disease transmission stock model for both versions. 4 chains, 100 warmup 100 samples, init radius 0.2. The resulting runs return mean values of gamma within 13.5% tolerance, mean values of beta within 63.75% tolerance, phi_inv within 17.5%, and phi to 181%. So, not exactly perfect reproducibility between the versions, but I then reran the live version and got something much much closer to the this-branch version, so I think it's within tolerance.

Have not tested the updated webR version, I'll give that a shot in a sec.

jsoules:
Installed the updated webR and reran data generation; no errors. (I think that's probably sufficient.)

jsoules (approved):

PR #363, opened 2026-06-10
Update .to_arviz for arviz base changes

WardBrian:
Arviz 1.0 had [many changes](https://python.arviz.org/en/stable/user_guide/migration_guide.html). This small PR hides the fact that our examples also got updated, see also https://gist.github.com/WardBrian/e47253bf29282d0eabf13616265d393e/revisions and https://gist.github.com/WardBrian/93d12876923790f23d9c5cb481e8cd34/revisions

### DIFF
--- a/gui/src/app/core/Scripting/pyodide/sp_load_draws.py
+++ b/gui/src/app/core/Scripting/pyodide/sp_load_draws.py
@@ -70,7 +70,9 @@ class DrawsObject:
         import arviz as az
 
         return az.from_dict(
-            posterior={pname: self.get(pname) for pname in self.parameter_names},
+            data={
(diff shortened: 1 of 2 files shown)


jsoules (approved):
Arviz isn't directly installed by us per se, it gets pulled in in our analysis scripts via Pyodide. Since we aren't pinning it, we hit a lot of interface changes when they went to v1.0 recently, hence this PR.

Confirmed the error behavior on the prod version. To reproduce, open one of the canned examples (I used Disease Transmission), run sampling, and attempt to run the default Python analysis script. Note that it fails with a complaint from arviz in our sp_load_draws() function.

Confirmed that this is fixed when running this branch.

Reviewed code changes here (they are minimal and unobjectionable) and those made to the gists for linear regression and disease transmission canned models.

PR #364, opened 2026-06-11
Version bumps

WardBrian:
Major versions have recently released of pyodide and emscripten.

The only breaking changes that affected us is that emscripten changed the default value of `FAKE_DYLIBS`, and pyodide now includes `lzma` by default.

How to verify:

- Run local compilation server, build examples
- Run data and analysis.py for disease transmission example

### DIFF
--- a/backend/local.mk
+++ b/backend/local.mk
@@ -12,9 +12,9 @@ LDLIBS_TBB ?= -ltbb
 TINYSTAN_SERIAL=true
 
 # could also uses -fexceptions which is more compatible, but slower
-CXXFLAGS+=-fwasm-exceptions
+CXXFLAGS+=-fwasm-exceptions # -sWASM_LEGACY_EXCEPTIONS=false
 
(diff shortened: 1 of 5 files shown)


backend/local.mk
```diff
@@ -12,9 +12,9 @@ LDLIBS_TBB ?= -ltbb
 TINYSTAN_SERIAL=true
 
 # could also uses -fexceptions which is more compatible, but slower
-CXXFLAGS+=-fwasm-exceptions
+CXXFLAGS+=-fwasm-exceptions # -sWASM_LEGACY_EXCEPTIONS=false
```
jsoules, line 15: any reason to keep the commented-out reference?

jsoules (approved):
No issues running the local compilation server. Had to merge the arviz changes to run disease-model analysis.py (unsurprisingly) but that done, everything works fine.

I'd say we're good here.

backend/local.mk
WardBrian: As a reminder to investigate at another time ;)

PR #371, opened 2026-07-22
Update monaco-editor

WardBrian:
Minor version bump, resolves a bunch of issues with `dompurify`


jsoules (approved):
Ran it up locally with the new version of Monaco.

As I understand it, we don't have any specific expected changed behavior, and I'm also not noticing any; this seems like it still runs, so I think we're fine to push it.

PR #378, opened 2026-09-22
Update to Stan 2.40

WardBrian:
Here's a model that uses one of the new functions:

```stan
data {
  int<lower=0> N;
  vector[N] x;
  vector[N] y;
}
parameters {
  real alpha;
  real beta;
  real<lower=0> sigma;
}
model {
  print(poisson_binomial_lpmf(1 | [0.6, 0.9]));
}
```

To test:

server:
```bash
cd backend && docker build -t stan-playground . && docker run -p 8083:8080 -e -it stan-playground 
```

frontend:
```
cd gui && yarn && yarn dev
```


jsoules (approved):
Executed the suggested example without problems.

I also reran the disease model example and did not notice a discrepancy with the existing published version.

